PR #32, opened 2024-12-07
Add Api impl for ListForwardedPayments.

G8XSU:
Based on #35

### DIFF
--- a/ldk-server/src/api/list_forwarded_payments.rs
+++ b/ldk-server/src/api/list_forwarded_payments.rs
@@ -1,0 +1,47 @@
+use crate::io::{
+	FORWARDED_PAYMENTS_PERSISTENCE_PRIMARY_NAMESPACE,
+	FORWARDED_PAYMENTS_PERSISTENCE_SECONDARY_NAMESPACE,
+};
+use crate::service::Context;
+use bytes::Bytes;
+use ldk_server_protos::api::{ListForwardedPaymentsRequest, ListForwardedPaymentsResponse};
+use ldk_server_protos::types::{ForwardedPayment, PageToken};
+use prost::Message;
+
+pub(crate) const LIST_FORWARDED_PAYMENTS_PATH: &str = "ListForwardedPayments";
+
+pub(crate) fn handle_list_forwarded_payments_request(
+	context: Context, request: ListForwardedPaymentsRequest,
+) -> Result<ListForwardedPaymentsResponse, ldk_node::NodeError> {
+	let page_token = request.page_token.map(|p| (p.token, p.index));
+	let list_response = context
+		.paginated_kv_store
+		.list(
+			FORWARDED_PAYMENTS_PERSISTENCE_PRIMARY_NAMESPACE,
+			FORWARDED_PAYMENTS_PERSISTENCE_SECONDARY_NAMESPACE,
+			page_token,
+		)
+		.map_err(|_| ldk_node::NodeError::ConnectionFailed)?;
+
+	let mut forwarded_payments: Vec<ForwardedPayment> = vec![];
+	for key in list_response.keys {
+		let forwarded_payment_bytes = context
+			.paginated_kv_store
+			.read(
+				FORWARDED_PAYMENTS_PERSISTENCE_PRIMARY_NAMESPACE,
+				FORWARDED_PAYMENTS_PERSISTENCE_SECONDARY_NAMESPACE,
+				&key,
+			)
+			.map_err(|_| ldk_node::NodeError::ConnectionFailed)?;
+		let forwarded_payment = ForwardedPayment::decode(Bytes::from(forwarded_payment_bytes))
+			.map_err(|_| ldk_node::NodeError::ConnectionFailed)?;
+		forwarded_payments.push(forwarded_payment);
+	}
+	let response = ListForwardedPaymentsResponse {
+		forwarded_payments,
+		next_page_token: list_response
+			.next_page_token
+			.map(|(token, index)| PageToken { token, index }),
+	};
+	Ok(response)
+}
--- a/ldk-server/src/main.rs
+++ b/ldk-server/src/main.rs
@@ -204,7 +204,7 @@ fn main() {
 					match res {
 						Ok((stream, _)) => {
 							let io_stream = TokioIo::new(stream);
-							let node_service = NodeService::new(Arc::clone(&node));
+							let node_service = NodeService::new(Arc::clone(&node), Arc::clone(&paginated_store) as Arc<dyn PaginatedKVStore + Send + Sync>);
 							runtime.spawn(async move {
 								if let Err(err) = http1::Builder::new().serve_connection(io_stream, node_service).await {
 									eprintln!("Failed to serve connection: {}", err);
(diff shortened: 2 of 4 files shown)


ldk-server/src/main.rs
```diff
@@ -97,22 +111,62 @@ fn main() {
...
+							total_fee_earned_msat,
+							skimmed_fee_msat,
+							claim_from_onchain_tx,
+							outbound_amount_forwarded_msat} => {
+
+							println!("PAYMENT_FORWARDED: with outbound_amount_forwarded_msat {}, total_fee_earned_msat: {}, inbound channel: {}, outbound channel: {}",
+								outbound_amount_forwarded_msat.unwrap_or(0), total_fee_earned_msat.unwrap_or(0), prev_channel_id, next_channel_id
+							);
+
+							let forwarded_payment = forwarded_payment_to_proto(payment_forwarded_event);
```
jkczyz, line 143: Do we need to use the proto version here? Seems it would better not to to the translation and allocate a vec.

ldk-server/src/service.rs
```diff
@@ -29,51 +25,66 @@ use crate::api::open_channel::{handle_open_channel, OPEN_CHANNEL_PATH};
...
 impl NodeService {
 	pub(crate) fn new(node: Arc<Node>) -> Self {
-		Self { node }
+		Self { context: Arc::new(Context { node }) }
 	}
 }
 
+pub(crate) struct Context {
+	pub(crate) node: Arc<Node>,
+}
```
jkczyz: Could you remind me why `PaginatedKvStore` can't be in `NodeService`? Is it just to save passing two parameters to the relevant handlers? If this is only need for one handler, then I'm not sure if we should bother adding the wrapper.

jkczyz: Hmm... seems less complicated if we don't add another struct. I don't see it as a circular dependency. The only reason why these functions aren't methods on `NodeService` is to get around the fact that `Service` uses `Future` instead of being an `async` trait. Using `Arc<NodeService>` is just like a `&self` parameter that works with `async`.

The wrapper is only needed because both `Service` and `Arc` don't belong to this crate, so we need another type to implement `Service` on. But that is only needed locally.

jkczyz: Here's the first part of the change. Then you just need to update the handlers to take `Arc<NodeService>` instead of `Arc<Node>`.

```diff
diff --git a/ldk-server/src/main.rs b/ldk-server/src/main.rs
index 6d31341..010a40f 100644
--- a/ldk-server/src/main.rs
+++ b/ldk-server/src/main.rs
@@ -3,7 +3,7 @@ mod io;
 mod service;
 mod util;
 
-use crate::service::NodeService;
+use crate::service::{ArcNodeService, NodeService};
 
 use ldk_node::{Builder, Event, LogLevel};
 
@@ -76,6 +76,8 @@ fn main() {
                node.config().listening_addresses.as_ref().unwrap().first().unwrap()
        );
 
+       let node_service = Arc::new(NodeService::new(Arc::clone(&node)));
+
        runtime.block_on(async {
                let mut sigterm_stream = match tokio::signal::unix::signal(SignalKind::terminate()) {
                        Ok(stream) => stream,
@@ -118,7 +120,7 @@ fn main() {
                                        match res {
                                                Ok((stream, _)) => {
                                                        let io_stream = TokioIo::new(stream);
-                                                       let node_service = NodeService::new(Arc::clone(&node));
+                                                       let node_service = ArcNodeService(Arc::clone(&node_service));
                                                        runtime.spawn(async move {
                                                                if let Err(err) = http1::Builder::new().serve_connection(io_stream, node_service).await {
                                                                        eprintln!("Failed to serve connection: {}", err);
diff --git a/ldk-server/src/service.rs b/ldk-server/src/service.rs
index 1062cbc..fa1c526 100644
--- a/ldk-server/src/service.rs
+++ b/ldk-server/src/service.rs
@@ -8,6 +8,7 @@ use hyper::{Request, Response, StatusCode};
 use prost::Message;
 
 use std::future::Future;
+use std::ops::Deref;
 use std::pin::Pin;
 use std::sync::Arc;
 
@@ -41,13 +42,15 @@ impl NodeService {
        }
 }
 
-impl Service<Request<Incoming>> for NodeService {
+pub(crate) struct ArcNodeService(pub Arc<NodeService>);
+
+impl Service<Request<Incoming>> for ArcNodeService {
        type Response = Response<Full<Bytes>>;
        type Error = hyper::Error;
        type Future = Pin<Box<dyn Future<Output = Result<Self::Response, Self::Error>> + Send>>;
 
        fn call(&self, req: Request<Incoming>) -> Self::Future {
-               let node = Arc::clone(&self.node);
+               let node = Arc::clone(&self.0.deref().node);
                // Exclude '/' from path pattern matching.
                match &req.uri().path()[1..] {
                        GET_NODE_INFO => Box::pin(handle_request(node, req, handle_get_node_info_request)),
@@ -94,7 +97,7 @@ async fn handle_request<
        F: Fn(Arc<Node>, T) -> Result<R, ldk_node::NodeError>,
 >(
        node: Arc<Node>, request: Request<Incoming>, handler: F,
-) -> Result<<NodeService as Service<Request<Incoming>>>::Response, hyper::Error> {
+) -> Result<<ArcNodeService as Service<Request<Incoming>>>::Response, hyper::Error> {
        // TODO: we should bound the amount of data we read to avoid allocating too much memory.
        let bytes = request.into_body().collect().await?.to_bytes();
        match T::decode(bytes) {
```

jkczyz: > Moreover node_service could have members which are not directly required for request-handling such as `authorizer`, and we wouldn't want to expose them as it is to handlers. Imo it would be better to have separation of concerns here.

We can always change it when that is the case. Handlers are simply methods on `NodeService`, IMO.

jkczyz: Not sure I really followed the example that you gave. Currently, `NodeService` is _only_ used for handling. Do you expect it to be used for something else (i.e., not handling) that would require an authorizer?

jkczyz: Was thinking about this more. Assuming an authenticator would be needed *before* handling, then it seems we can create `Context` inside `handle_request` rather than storing it in `NodeService`. Then it wouldn't even need to be an `Arc` since it could be passed by value to the handler. I'd be fine with that. As the code is currently written, a `NodeService` is composed of a `Context`, which doesn't seem right.

jkczyz: Right, inside `call` is accurate. I misspoke earlier.

jkczyz: Err... actually you need to create `Context` in `handle_request` to pass it to `call`, IIUC, since you need both the `Arc<Node>` and `Arc< PaginatedKVStore>` from `NodeService` to create `Context`.

jkczyz: Yeah, that looks good. Seems I was just mixing up the names for some reason. 😛

ldk-server/src/api/list_forwarded_payments.rs
```diff
@@ -0,0 +1,47 @@
...
+};
+use crate::service::Context;
+use bytes::Bytes;
+use ldk_server_protos::api::{ListForwardedPaymentsRequest, ListForwardedPaymentsResponse};
+use ldk_server_protos::types::{ForwardedPayment, PageToken};
+use prost::Message;
+
+pub(crate) const LIST_FORWARDED_PAYMENTS_PATH: &str = "ListForwardedPayments";
+
+pub(crate) fn handle_list_forwarded_payments_request(
```
jkczyz, line 13: Need to update service.rs to use these.

jkczyz (commented):
LGTM. Please squash

jkczyz (approved):

PR #46, opened 2025-02-26
Use LdkServerError for error handling in all APIs.

G8XSU:
Use LdkServerError for error handling in all APIs.

### DIFF
--- a/ldk-server-cli/src/main.rs
+++ b/ldk-server-cli/src/main.rs
@@ -1,6 +1,9 @@
 use clap::{Parser, Subcommand};
 use ldk_server_client::client::LdkServerClient;
 use ldk_server_client::error::LdkServerError;
+use ldk_server_client::error::LdkServerErrorCode::{
+	AuthError, InternalError, InternalServerError, InvalidRequestError, LightningError,
+};
 use ldk_server_client::ldk_server_protos::api::{
 	Bolt11ReceiveRequest, Bolt11SendRequest, Bolt12ReceiveRequest, Bolt12SendRequest,
 	GetBalancesRequest, GetNodeInfoRequest, ListChannelsRequest, ListPaymentsRequest,
@@ -123,7 +126,8 @@ async fn main() {
 					kind: Some(bolt11_invoice_description::Kind::Hash(hash)),
 				}),
 				(Some(_), Some(_)) => {
-					handle_error(LdkServerError::InternalError(
+					handle_error(LdkServerError::new(
+						InternalError,
 						"Only one of description or description_hash can be set.".to_string(),
 					));
 				},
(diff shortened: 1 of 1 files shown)


ldk-server/src/api/error.rs
```diff
@@ -52,3 +53,67 @@ impl fmt::Display for LdkServerErrorCode {
...
+			| NodeError::InvoiceRequestCreationFailed
+			| NodeError::OfferCreationFailed
+			| NodeError::RefundCreationFailed
+			| NodeError::PaymentSendingFailed
+			| NodeError::InvalidCustomTlvs
+			| NodeError::ProbeSendingFailed
+			| NodeError::ChannelCreationFailed
+			| NodeError::ChannelClosingFailed
+			| NodeError::ChannelConfigUpdateFailed
+			| NodeError::DuplicatePayment
```
jkczyz, line 96: Haven't looked through these exhaustively, but for BOLT11 `DuplicatePayment` is returned if the user tries to pay an invoice with the same payment hash twice. So shouldn't it map to an `InvalidRequestError`? Though for BOLT12 it would happen if we randomly generate the same payment id, which shouldn't happen.

ldk-server/src/api/onchain_send.rs
```diff
@@ -7,11 +9,16 @@ pub(crate) const ONCHAIN_SEND_PATH: &str = "OnchainSend";
...
+) -> Result<OnchainSendResponse, LdkServerError> {
 	let address = Address::from_str(&request.address)
 		.map_err(|_| ldk_node::NodeError::InvalidAddress)?
 		.require_network(context.node.config().network)
-		.map_err(|_| ldk_node::NodeError::InvalidAddress)?;
+		.map_err(|_| {
+			LdkServerError::new(
+				InvalidRequestError,
+				"Address is not valid for LdkServer's configured network.".to_string(),
+			)
```
jkczyz, line 20: Do we need use the mapping here in order to provide a more specific error?

ldk-server/src/api/get_payment_details.rs
```diff
@@ -8,9 +10,14 @@ pub(crate) const GET_PAYMENT_DETAILS_PATH: &str = "GetPaymentDetails";
...
 	context: Context, request: GetPaymentDetailsRequest,
-) -> Result<GetPaymentDetailsResponse, ldk_node::NodeError> {
-	let payment_id_bytes = <[u8; PaymentId::LENGTH]>::from_hex(&request.payment_id)
-		.map_err(|_| ldk_node::NodeError::InvalidPaymentId)?;
+) -> Result<GetPaymentDetailsResponse, LdkServerError> {
+	let payment_id_bytes =
+		<[u8; PaymentId::LENGTH]>::from_hex(&request.payment_id).map_err(|_| {
+			LdkServerError::new(
+				InvalidRequestError,
+				format!("Invalid payment_id, must be {}byte hex-string.", PaymentId::LENGTH),
```
jkczyz, line 18: a {}-byte

jkczyz: Note the dash, too.

ldk-server/src/api/onchain_send.rs
jkczyz: Yeah, I meant to say "not use" not "need use" lol.

jkczyz (commented):
LGTM. Please squash.

jkczyz (approved):

Issue #37, opened 2024-12-12
Events subscription


dzdidi:
Allow (via web-sockets?) to subscribe to events:
- payment state update
- channel state update
- forward state update
- invoice state update
- request state update
- peer state update

The more gradual events the better, for example in case of channel possible state transition could be:
`openning -> open -> closing -> closed`
as each state update can have specific business logic attached to it

Issue #38, opened 2024-12-12
Telemetry


dzdidi:
It is desired to have certain amount of insight into how service operates historically and at any moment of time for both business and operational intelligence.

Prometheus metrics for different stats like:
- balances in different states and layers
- forward payments count
- forward fees earned
- number of channels in different states (public  / private)
- inbound/outbound capacity 
- etc

arturgontijo:
Is anybody working on this one?

PR #51, opened 2025-03-12
Introduce EventPublisher trait.

G8XSU:
The LDK Server needs a flexible and reliable mechanism to publish informational events (e.g., transaction updates, channel status changes) to external messaging systems for integration with other external services in a decoupled manner.

To achieve this, we introduce the EventPublisher trait, which defines a trait for event publishing. This allows the daemon to support multiple messaging backends via feature flags.

Since underlying messaging systems are expected to support durable buffering, this keeps the LDK Server decoupled from event storage and handling, while enabling multiple consumers to process events independently.

This is in contrast with in-memory event queues, polling, or WebSockets, since they either lack durability, tightly couple the daemon to consumers, or require constant connectivity. In-memory queues risk losing events on restart, polling burdens the daemon with service calls from each consumer, and WebSockets demand real-time client availability—none of which efficiently support multiple consumers or can scale effectively.

Note: Feature flags enable optional messaging backends, minimizing dependencies and avoiding restriction to a specific messaging system for users.

As part of #37

### DIFF
--- a/ldk-server/src/main.rs
+++ b/ldk-server/src/main.rs
@@ -13,6 +13,7 @@ use tokio::signal::unix::SignalKind;
 use hyper::server::conn::http1;
 use hyper_util::rt::TokioIo;
 
+use crate::io::events::event_publisher::{EventPublisher, NoopEventPublisher};
 use crate::io::persist::paginated_kv_store::PaginatedKVStore;
 use crate::io::persist::sqlite_store::SqliteStore;
 use crate::io::persist::{
@@ -26,6 +27,8 @@ use hex::DisplayHex;
 use ldk_node::config::Config;
 use ldk_node::lightning::ln::channelmanager::PaymentId;
 use ldk_node::logger::LogLevel;
+use ldk_server_protos::events;
+use ldk_server_protos::events::{event_envelope, EventEnvelope};
 use prost::Message;
 use rand::Rng;
 use std::fs;
@@ -97,6 +100,8 @@ fn main() {
 				std::process::exit(-1);
 			},
 		});
+
+	let event_publisher: Arc<dyn EventPublisher> = Arc::new(NoopEventPublisher);
 
 	println!("Starting up...");
 	match node.start_with_runtime(Arc::clone(&runtime)) {
@@ -199,6 +204,18 @@ fn main() {
 							rand::thread_rng().fill(&mut forwarded_payment_id);
 
 							let forwarded_payment_creation_time = SystemTime::now().duration_since(UNIX_EPOCH).expect("Time must be > 1970").as_secs() as i64;
+
+							match event_publisher.publish(EventEnvelope {
+									event: Some(event_envelope::Event::PaymentForwarded(events::PaymentForwarded {
+											forwarded_payment: Some(forwarded_payment.clone()),
+									})),
+							}).await {
+								Ok(_) => {},
+								Err(e) => {
+									println!("Failed to publish 'PaymentForwarded' event: {}", e);
+									continue;
+								}
+							};
 
 							match paginated_store.write(FORWARDED_PAYMENTS_PERSISTENCE_PRIMARY_NAMESPACE,FORWARDED_PAYMENTS_PERSISTENCE_SECONDARY_NAMESPACE,
 								&forwarded_payment_id.to_lower_hex_string(),


ldk-server/src/io/events/event_publisher.rs
```diff
@@ -0,0 +1,40 @@
...
+/// The underlying messaging system is expected to support durably buffered events,
+/// enabling easy decoupling between the LDK Server and event consumers.
+#[async_trait]
+pub trait EventPublisher {
+	/// Publishes an event to the underlying messaging system.
+	///
+	/// # Arguments
+	/// * `event` - The event message to publish, provided as a `Bytes` buffer.
+	///             This contains a serialized protobuf message as defined in [`ldk_server_protos::events`].
+
```
jkczyz, line 26: nit: Missing `///`

ldk-server/src/io/events/event_publisher.rs
```diff
@@ -0,0 +1,40 @@
...
+/// with specific implementations enabled via feature flags.
+///
+/// Events are serialized Protocol Buffers ([protobuf](https://protobuf.dev/)) objects/messages, defined in
+/// [`ldk_server_protos::events`]. The `Bytes` argument represents the serialized form of these
+/// protobuf messages. Consumers can use the defined protobufs for deserialization.
+///
+/// The underlying messaging system is expected to support durably buffered events,
+/// enabling easy decoupling between the LDK Server and event consumers.
+#[async_trait]
+pub trait EventPublisher {
```
jkczyz, line 20: Could you provide a rough idea of how LDK Server will be parameterized by this and how it will be used?

ldk-server-protos/src/proto/events.proto
```diff
@@ -0,0 +1,22 @@
...
+
+// Represents different types of events, useful for routing and filtering.
+enum EventType {
+  UNKNOWN = 0;
+  PAYMENT_FORWARDED = 1;
+}
+
+// EventEnvelope wraps different event types in a single message to be used by EventPublisher.
+message EventEnvelope {
+  EventType event_type = 1;
```
jkczyz, line 13: Why do we need `EventType` if we are using `oneof`?

ldk-server-protos/src/proto/events.proto
```diff
@@ -0,0 +1,22 @@
+syntax = "proto3";
+import "types.proto";
+package events;
+
+// Represents different types of events, useful for routing and filtering.
```
jkczyz, line 5: Not sure what is meant by "useful for routing and filtering".

ldk-server/src/io/events/event_publisher.rs
jkczyz: Ah, ok, I was mistakenly thinking this was related to #38.

ldk-server/src/main.rs
```diff
@@ -200,6 +205,19 @@ fn main() {
...
+									event: Some(event_envelope::Event::PaymentForwarded(events::PaymentForwarded {
+											forwarded_payment: Some(forwarded_payment.clone()),
+									})),
+							}).await{
+								Ok(_) => {},
+								Err(e) => {
+									println!("Failed to publish 'PaymentForwarded' event: {}", e);
+									return;
+								}
+							};
```
jkczyz, line 219: Below, if we fail writing to storage, we wouldn't call `event_node.event_handled()`. Thus, on the next iteration we would retry the event, IIUC, causing it to be published twice.

ldk-server/src/io/events/event_publisher.rs
```diff
@@ -0,0 +1,57 @@
...
+	/// In order to ensure no events are lost, implementors of this trait must publish events
+	/// durably to underlying messaging system. An event is considered published when
+	/// [`EventPublisher::publish`] returns `Ok(())`, thus implementors MUST durably persist/publish events *before*
+	/// returning `Ok(())`.
+	///
+	/// # Errors
+	/// May return an [`LdkServerErrorCode::InternalServerError`] if the event cannot be published,
+	/// such as due to network failures, misconfiguration, or transport-specific issues.
+	/// If event publishing fails, the LDK Server will retry publishing the event indefinitely, which
+	/// may degrade performance until the underlying messaging system is operational again.
```
jkczyz, line 40: Doesn't the return on line 217 of main.rs mean publishing is *not* retried?

ldk-server/src/main.rs
jkczyz: > Yes, it is a common pattern when both event publishing and db save has to happen in same operation. Event publishing happens before db save and duplicate events are generally expected. (even the underlying messaging systems can create duplicate events to guarantee delivery.)

Ah, is the `EventPublisher` implementation expected to de-duplicate for subscribers?

> We would ideally want some kind of circuit-breaker or delay here in case our event-queue or database are temporarily unavailable.

Do you mean a retry or failure? IIUC, seems like we currently fail if the event-queue is unavailable and retry without delay if the database isn't.

ldk-server-protos/src/proto/events.proto
jkczyz: Do you mean they would decided whether or not whether to deserialize the field based on `event_type`?  Can't the protobuf serve that purpose?

ldk-server-protos/src/proto/events.proto
jkczyz: Yeah, I guess I don't understand why switching / matching on the `oneof` can't be used. Seems like `event_type` is duplicating that.

jkczyz (commented):
LGTM. Please squash.

PR #53, opened 2025-03-13
Add RabbitMQ-based Implementation of EventPublisher Trait

G8XSU:
Note: It is only enabled if we enable `events-rabbitmq` feature.

RabbitMQ was selected because it is one of the most robust and battle-tested open-source messaging systems, allowing for durable, buffered messages and flexible exchange options.

This ensures that events are not lost, even in the case of restarts, network failures, or consumer failures, while keeping consumers decoupled.

* Feature-based: Only enabled if we enable `events-rabbitmq` feature, this allows us to support multiple implementations and no strict coupling with a single impl while keep our dependencies to the minimum.
* Durable Publishing: Uses publisher confirms and persistent delivery mode to guarantee events are stored by RabbitMQ before returning Ok(()).
* Lazy Initialized connection: Wraps Connection and Channel in Option for lazy connection init via ensure_connected.
* Reconnection Handling: Implements a single reconnect attempt per publish call. Event publishing is already retried, so this keeps it simple for now.

Depends on #51 
As part of #37

### DIFF
--- a/ldk-server/Cargo.toml
+++ b/ldk-server/Cargo.toml
@@ -25,3 +25,9 @@ lapin = { version = "2.4.0", features = ["rustls"], default-features = false, op
 [features]
 default = []
 events-rabbitmq = ["dep:lapin"]
+
+# Feature-flags related to integration tests.
+integration-tests-events-rabbitmq = ["events-rabbitmq"]
+
+[dev-dependencies]
+futures-util = "0.3.31"
--- a/ldk-server/src/io/events/rabbitmq/mod.rs
+++ b/ldk-server/src/io/events/rabbitmq/mod.rs
@@ -142,3 +142,91 @@ impl EventPublisher for RabbitMqEventPublisher {
 		}
 	}
 }
+
+#[cfg(test)]
+#[cfg(feature = "integration-tests-events-rabbitmq")]
+mod integration_tests_events_rabbitmq {
+	use super::*;
+	use lapin::{
+		options::{BasicAckOptions, BasicConsumeOptions, QueueBindOptions, QueueDeclareOptions},
+		types::FieldTable,
+		Channel, Connection,
+	};
+	use ldk_server_protos::events::event_envelope::Event;
+	use ldk_server_protos::events::PaymentForwarded;
+	use std::io;
+	use std::time::Duration;
+	use tokio;
+
+	use futures_util::stream::StreamExt;
+	#[tokio::test]
+	async fn test_publish_and_consume_event() {
+		let config = RabbitMqConfig {
+			connection_string: "amqp://guest:guest@localhost:5672/%2f".to_string(),
+			exchange_name: "test_exchange".to_string(),
+		};
+
+		let publisher = RabbitMqEventPublisher::new(config.clone());
+
+		let conn = Connection::connect(&config.connection_string, ConnectionProperties::default())
+			.await
+			.expect("Failed make rabbitmq connection");
+		let channel = conn.create_channel().await.expect("Failed to create rabbitmq channel");
+
+		let queue_name = "test_queue";
+		setup_queue(&queue_name, &channel, &config).await;
+
+		let event =
+			EventEnvelope { event: Some(Event::PaymentForwarded(PaymentForwarded::default())) };
+		publisher.publish(event.clone()).await.expect("Failed to publish event");
+
+		consume_event(&queue_name, &channel, &event).await.expect("Failed to consume event");
+	}
+
+	async fn setup_queue(queue_name: &str, channel: &Channel, config: &RabbitMqConfig) {
+		channel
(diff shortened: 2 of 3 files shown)


ldk-server/ldk-server.config
```diff
@@ -14,9 +14,10 @@
...
-  // Bitcoin Core's RPC user.
-  "bitcoind_rpc_user": "polaruser",
+  // RabbitMQ connection string. (only required if using RabbitMQ based events using `events-rabbitmq` feature)
+  "rabbitmq_connection_string": "",
+
+  // RabbitMQ exchange name. (only required if using RabbitMQ based events using `events-rabbitmq` feature)
+  "rabbitmq_exchange_name": ""
 
-  // Bitcoin Core's RPC password.
-  "bitcoind_rpc_password": "polarpass"
```
jkczyz, line 21: Why remove the bitcoind config parameters?

ldk-server/Cargo.toml
```diff
@@ -18,3 +18,17 @@ hex = { package = "hex-conservative", version = "0.2.1", default-features = fals
...
+[features]
+default = []
+events-rabbitmq = ["dep:lapin"]
+
+
+# Feature-flags related to integration tests.
+integration-tests-events-rabbitmq = ["events-rabbitmq"]
+
+[dev-dependencies]
+futures-util = "0.3.31"
```
jkczyz, line 34: Missing newline

ldk-server/Cargo.toml
```diff
@@ -18,3 +18,17 @@ hex = { package = "hex-conservative", version = "0.2.1", default-features = fals
...
 async-trait = { version = "0.1.85", default-features = false }
+
+# Required for RabittMQ based EventPublisher. Only enabled for `events-rabbitmq` feature.
+lapin = { version = "2.5.1", optional = true }
+
+[features]
+default = []
+events-rabbitmq = ["dep:lapin"]
+
+
```
jkczyz, line 29: Extra line

ldk-server/Cargo.toml
```diff
@@ -18,3 +18,17 @@ hex = { package = "hex-conservative", version = "0.2.1", default-features = fals
 rusqlite = { version = "0.31.0", features = ["bundled"] }
 rand = { version = "0.8.5", default-features = false }
 async-trait = { version = "0.1.85", default-features = false }
+
+# Required for RabittMQ based EventPublisher. Only enabled for `events-rabbitmq` feature.
+lapin = { version = "2.5.1", optional = true }
```
jkczyz, line 23: Looks like this is require a change of the MSRV to 1.81.0 with the feature enabled.

ldk-server/src/io/events/rabbitmq/mod.rs
```diff
@@ -0,0 +1,238 @@
...
+		}
+	}
+}
+
+// Ensure the publisher is Send and Sync for use in async contexts
+unsafe impl Send for RabbitMqEventPublisher {}
+unsafe impl Sync for RabbitMqEventPublisher {}
+
+#[cfg(test)]
+#[cfg(feature = "integration-tests-events-rabbitmq")]
```
jkczyz, line 151: We'll need to update CI to run these.

ldk-server/src/main.rs
```diff
@@ -103,6 +105,15 @@ fn main() {
...
 	let event_publisher: Arc<dyn EventPublisher> = Arc::new(NoopEventPublisher);
 
+	#[cfg(feature = "events-rabbitmq")]
+	let event_publisher: Arc<dyn EventPublisher> = {
+		let rabbitmq_config = RabbitMqConfig {
+			connection_string: config_file.rabbitmq_connection_string,
+			exchange_name: config_file.rabbitmq_exchange_name,
+		};
+		Arc::new(RabbitMqEventPublisher::new(rabbitmq_config))
+	};
```
jkczyz, line 115: Should we open a connection to ensure the config parameters are valid?

jkczyz: Makes sense, though how would an operator be aware of the rabbitmq unavailability? IIUC, unavailability would cause the event queue to be backed up. The underlying node would still progress, but from the user's perspective the payment store would be stale since event publishing failure would prevent persisting the payment.

This doesn't need to hold up the PR, but just wondering how this would be monitored.

ldk-server/src/io/events/rabbitmq/mod.rs
```diff
@@ -143,6 +143,90 @@ impl EventPublisher for RabbitMqEventPublisher {
 	}
 }
 
-// Ensure the publisher is Send and Sync for use in async contexts
-unsafe impl Send for RabbitMqEventPublisher {}
-unsafe impl Sync for RabbitMqEventPublisher {}
```
jkczyz: Should be removed from the first commit.

jkczyz (commented):
LGTM

jkczyz (approved):

PR #57, opened 2025-03-18
Start publishing more payment events using EventPublisher.

G8XSU:
Publishes events for:
* PaymentReceived
* PaymentSuccessful
* PaymentFailed

As part of #37

### DIFF
--- a/ldk-server/src/main.rs
+++ b/ldk-server/src/main.rs
@@ -14,6 +14,7 @@ use hyper::server::conn::http1;
 use hyper_util::rt::TokioIo;
 
 use crate::io::events::event_publisher::{EventPublisher, NoopEventPublisher};
+use crate::io::events::get_event_name;
 use crate::io::persist::paginated_kv_store::PaginatedKVStore;
 use crate::io::persist::sqlite_store::SqliteStore;
 use crate::io::persist::{
@@ -29,12 +30,14 @@ use ldk_node::lightning::ln::channelmanager::PaymentId;
 use ldk_node::logger::LogLevel;
 use ldk_server_protos::events;
 use ldk_server_protos::events::{event_envelope, EventEnvelope};
+use ldk_server_protos::types::Payment;
 use prost::Message;
 use rand::Rng;
 use std::fs;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
 use std::time::{SystemTime, UNIX_EPOCH};
+use tokio::select;
 
 const USAGE_GUIDE: &str = "Usage: ldk-server <config_path>";
 
@@ -124,14 +127,14 @@ fn main() {
 			Err(e) => {
 				println!("Failed to register for SIGTERM stream: {}", e);
 				std::process::exit(-1);
-			},
+			}
 		};
 		let event_node = Arc::clone(&node);
 		let rest_svc_listener = TcpListener::bind(config_file.rest_service_addr)
 			.await
 			.expect("Failed to bind listening port");
 		loop {
-			tokio::select! {
+			select! {
 				event = event_node.next_event_async() => {
 					match event {
 						Event::ChannelPending { channel_id, counterparty_node_id, .. } => {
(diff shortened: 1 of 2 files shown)


ldk-server/src/main.rs
```diff
@@ -154,18 +156,44 @@ fn main() {
 								payment_id, payment_hash, amount_msat
 							);
 							let payment_id = payment_id.expect("PaymentId expected for ldk-server >=0.1");
-							upsert_payment_details(&event_node, Arc::clone(&paginated_store) as Arc<dyn PaginatedKVStore>, &payment_id);
+
+							publish_event_and_upsert_payment(&payment_id,
+								|payment| event_envelope::Event::PaymentReceived(events::PaymentReceived {
+                    payment: Some(payment.clone()),
+                }),
```
jkczyz, line 163: Indentation off here and below. Looks to be using spaces instead of tabs?

ldk-server/src/main.rs
```diff
@@ -154,18 +156,44 @@ fn main() {
 								payment_id, payment_hash, amount_msat
 							);
 							let payment_id = payment_id.expect("PaymentId expected for ldk-server >=0.1");
-							upsert_payment_details(&event_node, Arc::clone(&paginated_store) as Arc<dyn PaginatedKVStore>, &payment_id);
+
+							publish_event_and_upsert_payment(&payment_id,
+								|payment| event_envelope::Event::PaymentReceived(events::PaymentReceived {
+                    payment: Some(payment.clone()),
```
jkczyz, line 162: Why clone here if `payment` is passed by value?

ldk-server/src/main.rs
```diff
@@ -266,30 +294,48 @@ fn main() {
...
-				event_node.event_handled();
-			},
+
+		let event = payment_to_event(payment.clone());
+		let event_name = get_event_name(&event);
+		match event_publisher.publish(EventEnvelope { event: Some(event) }).await {
+			Ok(_) => {},
 			Err(e) => {
-				eprintln!("Failed to write payment to persistence: {}", e);
+				println!("Failed to publish '{}' event: {}", event_name, e);
```
jkczyz, line 310: Would it be worth providing more information other than the event name? Does it alone even give us any useful information? Maybe include the payment id?

ldk-server/src/main.rs
```diff
@@ -154,18 +156,44 @@ fn main() {
...
+								&event_node,
+								Arc::clone(&event_publisher),
+								Arc::clone(&paginated_store)).await;
 						},
 						Event::PaymentClaimable {payment_id, ..} => {
-							upsert_payment_details(&event_node, Arc::clone(&paginated_store) as Arc<dyn PaginatedKVStore>, &payment_id);
+							if let Some(payment_details) = event_node.payment(&payment_id) {
+								let payment = payment_to_proto(payment_details);
+								upsert_payment_details(&event_node, Arc::clone(&paginated_store), &payment);
+							} else{
```
jkczyz, line 194: nit: missing space

ldk-server/src/main.rs
```diff
@@ -154,18 +156,44 @@ fn main() {
...
+								|payment| event_envelope::Event::PaymentFailed(events::PaymentFailed {
+                    payment: Some(payment.clone()),
+                }),
+								&event_node,
+								Arc::clone(&event_publisher),
+								Arc::clone(&paginated_store)).await;
 						},
 						Event::PaymentClaimable {payment_id, ..} => {
-							upsert_payment_details(&event_node, Arc::clone(&paginated_store) as Arc<dyn PaginatedKVStore>, &payment_id);
+							if let Some(payment_details) = event_node.payment(&payment_id) {
```
jkczyz, line 191: Could we be consistent with `publish_event_and_upsert_payment` and do the lookup check inside `upsert_payment_details`?

ldk-server/src/main.rs
jkczyz: nit: Could have just not cloned in the closures, FWIW. Cloning at the call site would be better IMO given it is a conversion function and you needed the value again later.

jkczyz (approved):
LGTM. Please squash

jkczyz (approved):

PR #71, opened 2025-10-17
Add docker

moisesPompilio:
Adds **Docker** support for building and testing **LDK Server**. Includes a multi-stage **Dockerfile** for optimized builds and a **docker-compose.yml** for a full testing environment with **Bitcoin Core, CLN, and LND**.

1. **Dockerfile**: Multi-stage build with builder stage for Rust compilation (supports `BUILD_FEATURES`) and final stage for runtime. Exposes ports **3000** and **3001** for REST and Lightning connectivity.
2. **docker-compose.yml**: Sets up a complete testing environment with **Bitcoin Core in regtest mode**, **CLN** (Core Lightning), and **LND** for Lightning Network interactions.
  -  - **Default** profile: Runs Bitcoin Core, CLN, LND (`docker compose up`).
 -  - **ldk-server** profile: Runs Bitcoin Core, CLN, LND, **LDK Server** (`docker compose --profile ldk-server up`).
  -  - **rabbitmq** profile: Runs Bitcoin Core, CLN, LND, **RabbitMQ** (`docker compose --profile rabbitmq up`).

this pr depends on https://github.com/lightningdevkit/ldk-server/pull/69

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,123 @@
+services:
+  ldk-server:
+    container_name: ldk-server
+    build:
+      context: .
+      args:
+        BUILD_FEATURES: ""
+    command: [
+      "--node-listening-address=0.0.0.0:3001",
+      "--node-rest-service-address=0.0.0.0:3000",
+      "--bitcoind-rpc-host=ldk-server-bitcoin",
+      "--bitcoind-rpc-port=18443",
+      "--bitcoind-rpc-user=user",
+      "--bitcoind-rpc-password=pass",
+    ]
+    ports:
+      - "3000:3000"
+      - "3001:3001"
+    networks:
+      - ldk-server
+    depends_on:
+      bitcoin:
+        condition: service_healthy
+    profiles: ["ldk-server"]
+
+  bitcoin:
+    container_name: ldk-server-bitcoin
+    image: blockstream/bitcoind:30.0@sha256:e1351bb0d735893abee13af422d4209069592d8113f8b0aa5ed1458681bfb715
+    platform: linux/amd64
+    command:
+      [
+        "bitcoind",
+        "-printtoconsole",
+        "-regtest=1",
+        "-rpcallowip=0.0.0.0/0",
+        "-rpcbind=0.0.0.0",
+        "-rpcuser=user",
+        "-rpcpassword=pass",
+        "-fallbackfee=0.00001",
+        "-zmqpubrawblock=tcp://0.0.0.0:28332",
+        "-zmqpubrawtx=tcp://0.0.0.0:28333"
+      ]
+    ports:
+      - "18443:18443"
+      - "18444:18444"
+    networks:
+      - ldk-server
+    healthcheck:
+      test: ["CMD", "bitcoin-cli", "-regtest", "-rpcuser=user", "-rpcpassword=pass", "getblockchaininfo"]
+      interval: 5s
+      timeout: 10s
+      retries: 5
+      # Example command: bitcoin-cli -regtest -rpcuser=user -rpcpassword=pass getblockchaininfo
+
+  lnd:
+    image: lightninglabs/lnd:v0.18.5-beta@sha256:2b560c9beb559c57ab2f2da1dfed80d286cf11a6dc6e4354cab84aafba79b6f6
+    container_name: ldk-server-lnd
+    ports:
+      - "9735:9735"   # P2P
+      - "10009:10009" # RPC
+    command:
+      - "--noseedbackup"
+      - "--trickledelay=5000"
+      - "--alias=ldk-node-lnd-test"
+      - "--externalip=lnd:9735"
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -0,0 +1,122 @@
...
+    networks:
+      - ldk-server
+    depends_on:
+      bitcoin:
+        condition: service_healthy
+    profiles: ["ldk-server"]
+
+  bitcoin:
+    container_name: ldk-server-bitcoin
+    image: blockstream/bitcoind:29.1@sha256:9fcffa83feed0fc382dfb2f26990ca07656d150ba49434096a5aeedac6695a01
```
tnull: Would be cool to go for v30 directly here, just pinged somebody from Blockstream to check if they intend to upgrade the Docker image soon.

docker-compose.yml
```diff
@@ -0,0 +1,122 @@
...
+      - "19846:19846"  # RPC
+      - "9736:9736"    # P2P
+    depends_on:
+      bitcoin:
+        condition: service_healthy
+    networks:
+      - ldk-server
+      # Example command: lightning-cli --network=regtest getinfo
+
+  rabbitmq:
```
tnull: Not entirely convinced we need this right now, but maybe better to have it.

ldk-server/ldk-server-config.toml
```diff
@@ -2,7 +2,7 @@
 [node]
 network = "regtest"                           # Bitcoin network to use
 listening_address = "localhost:3001"          # Lightning node listening address
-rest_service_address = "127.0.0.1:3002"       # LDK Server REST address
+rest_service_address = "127.0.0.1:3000"       # LDK Server REST address
```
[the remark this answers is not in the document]
moisesPompilio: My main motivation is that `ldk-server-cli` defaults to port `3000` for the RPC. If this config file uses `3002`, users would always need to manually pass `-b localhost:3002`, which isn’t convenient, especially when running it through Docker where the config file is used directly. I believe `3002` was originally chosen because `3000` conflicted with [Polar](https://github.com/jamaljsr/polar), which uses it by default, but using `3002` could also cause conflicts with Esplora in the future once support for it is added.

docker-compose.yml
moisesPompilio: I added this command example to make it easier for anyone working with these containers, so they don’t have to look up the basic commands each time and everything stays consistent with the config.

docker-compose.yml
tnull: And there it is: https://hub.docker.com/r/blockstream/bitcoind

Please update!

docker-compose.yml
```diff
@@ -0,0 +1,123 @@
+services:
+  ldk-server:
+    container_name: ldk-server
+    build:
+      context: .
+      args:
+        BUILD_FEATURES: ""
+    command: [
+      "--node-listening-address=0.0.0.0:3001",
```
benthecarman: seems weird for ldk-server we use the non-default 3001 but give the default 9735 to lnd

Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+    fi
+
+FROM debian:13.1-slim@sha256:1caf1c703c8f7e15dcf2e7769b35000c764e6f50e4d7401c355fb0248f3ddfdb
+
+COPY --from=builder /opt/app/target/release/ldk-server /usr/local/bin/ldk-server
+COPY --from=builder /opt/app/target/release/ldk-server-cli /usr/local/bin/ldk-server-cli
+COPY --from=builder  /opt/app/ldk-server/ldk-server-config.toml /usr/local/bin/ldk-server-config.toml
+RUN chmod +x /usr/local/bin/ldk-server
+
+EXPOSE 3000 3001
```
benthecarman: should expose 9735

Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+  build-essential \
+  cmake \
+  curl \
+  git \
+  libssl-dev \
+  pkg-config
+
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+ENV PATH="/root/.cargo/bin:${PATH}"
+RUN rustup default 1.90.0
```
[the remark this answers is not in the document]
moisesPompilio: Makes sense, I replaced Debian with Rustup instead.

docker-compose.yml
moisesPompilio: LND uses port 9735, and LDK uses port 3001. I checked in my tests, and LDK is indeed using 3001.

Dockerfile
benthecarman: the dockerfile doesn't contain lnd, only ldk-server

benthecarman: The `docker-compose` here will just be an example regtest lightning network people will set up for testing, but the `Dockerfile` is actually how people will deploy production nodes so we should be optimizing for that.

PR #72, opened 2025-10-30
Bump `ldk-node` dependency

jkczyz:
Bump the `ldk-node` dependency to support splicing.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,16 +9,16 @@ jobs:
         platform: [
           ubuntu-latest,
           macos-latest,
-          ]
-        toolchain: 
+        ]
+        toolchain:
           [ stable,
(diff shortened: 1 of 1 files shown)


Cargo.lock
tnull: Seems we'll need to check in a separate `Cargo-minimal.lock` file that we use for MSRV CI.

jkczyz: Ah, I usually run `cargo` with our MSRV, but hadn't done so here. Should I just build using `cargo +1.75.0` for any `Cargo.lock` updates. Not sure what the best practices are.

tnull (dismissed):

benthecarman:
> > LGTM, though it seems I don't have approval power.

since its your PR, you can't approve it

PR #78, opened 2025-12-09
Make cli output json

benthecarman:
Previously we would output the debug format of each of our return times which worked, but was not very usable. You could not easily pipe results with cli tools or anything like that. This changes the results to be output to json to make it easier to use and more similiar to what people expect.

This had one annoyance because since our structs are generated using prost, we cannot just simply add the serde traits to the structs, so instead we recreate the types we are going to output in the cli and covert them. There are a couple places where I just created "empty" default versions of types because we had to handle the `None` case because of how proto handles structs, not totally sure if this is the best move or if we should do something like an `unreachable` instead

### DIFF
--- a/ldk-server-cli/Cargo.toml
+++ b/ldk-server-cli/Cargo.toml
@@ -4,7 +4,8 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-ldk-server-client = { path = "../ldk-server-client" }
+ldk-server-client = { path = "../ldk-server-client", features = ["serde"] }
 clap = { version = "4.0.5", default-features = false, features = ["derive", "std", "error-context", "suggestions", "help"] }
 tokio = { version = "1.38.0", default-features = false, features = ["rt-multi-thread", "macros"] }
-prost = { version = "0.11.6", default-features = false}
+serde = "1.0"
+serde_json = "1.0"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -546,6 +546,9 @@ name = "bytes"
 version = "1.10.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d71b6127be86fdcfddb610f7182ac57211d4b18a3e9c82eb2d17662f2227ad6a"
+dependencies = [
+ "serde",
+]
 
 [[package]]
 name = "cbc"
@@ -1739,7 +1742,8 @@ version = "0.1.0"
 dependencies = [
  "clap",
  "ldk-server-client",
- "prost",
+ "serde",
+ "serde_json",
  "tokio",
 ]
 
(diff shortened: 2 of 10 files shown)


ldk-server-cli/Cargo.toml
```diff
@@ -7,4 +7,6 @@ edition = "2021"
 ldk-server-client = { path = "../ldk-server-client" }
 clap = { version = "4.0.5", default-features = false, features = ["derive", "std", "error-context", "suggestions", "help"] }
 tokio = { version = "1.38.0", default-features = false, features = ["rt-multi-thread", "macros"] }
-prost = { version = "0.11.6", default-features = false}
```
tnull: Hmm, could be debated if we want to deviate from the 'protobuf-over-HTTP' approach, but I'm not sure if we should introduce all the boilerplate to support *yet another* encoding of the same request/response data?

tnull: Okay, but I have to say I still hate the idea of taking on the maintainance burden of almost 1000 LoC *just* to be able to print as valid JSON. If that's important we should really find another way to do it rather than just having an agent spit out a lot of boilerplate that nobody ever wants to read/maintain.

Turns out it's really easy to add serde/JSON support on the generated protos by just adding:

```diff
diff --git a/ldk-server-protos/Cargo.toml b/ldk-server-protos/Cargo.toml
index 5d16484..65894e2 100644
--- a/ldk-server-protos/Cargo.toml
+++ b/ldk-server-protos/Cargo.toml
@@ -7,6 +7,8 @@ build = "build.rs"

 [dependencies]
 prost = { version = "0.11.6", default-features = false, features = ["std", "prost-derive"] }
+serde = { version = "1.0", features = ["derive"] }
+bytes = { version = "1", features = ["serde"] }

 [target.'cfg(genproto)'.build-dependencies]
 prost-build = { version = "0.11.6" , default-features = false}
diff --git a/ldk-server-protos/build.rs b/ldk-server-protos/build.rs
index 76eb77a..ac5d0ce 100644
--- a/ldk-server-protos/build.rs
+++ b/ldk-server-protos/build.rs
@@ -13,6 +13,8 @@ fn main() {
 #[cfg(genproto)]
 fn generate_protos() {
        prost_build::Config::new()
+               .type_attribute(".", "#[derive(serde::Serialize, serde::Deserialize)]")
+               .type_attribute(".", "#[serde(rename_all = \"snake_case\")]")
                .bytes(&["."])
                .compile_protos(
                        &[
```

Also see https://github.com/tokio-rs/prost/issues/75 and https://protobuf.dev/programming-guides/proto3/#json

benthecarman: Okay good find, much easier

PR #79, opened 2025-12-09
Allow to set `SendingParameters` in BOLT11 & BOLT12 

### LINKED ISSUE #4: Allow to set SendingParameters in BOLT11 & BOLT12 send

Anyitechs:
This PR introduces the ability to set `SendingParameters` in BOLT11 and BOLT12 send methods.

Closes #4

### DIFF
--- a/ldk-server/src/api/bolt11_send.rs
+++ b/ldk-server/src/api/bolt11_send.rs
@@ -1,5 +1,7 @@
 use crate::api::error::LdkServerError;
+use crate::api::error::LdkServerErrorCode::InvalidRequestError;
 use crate::service::Context;
+use ldk_node::lightning::routing::router::RouteParametersConfig;
 use ldk_node::lightning_invoice::Bolt11Invoice;
 use ldk_server_protos::api::{Bolt11SendRequest, Bolt11SendResponse};
 use std::str::FromStr;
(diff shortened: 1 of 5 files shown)


ldk-server-protos/src/proto/types.proto
```diff
@@ -683,3 +683,24 @@ message Bolt11InvoiceDescription {
...
+  // The maximum total fees, in millisatoshi, that may accrue during route finding.
+  // Defaults to 1% of the payment amount + 50 sats
+  optional uint64 max_total_routing_fee_msat = 1;
+
+  // The maximum total CLTV delta we accept for the route.
+  // Defaults to [`DEFAULT_MAX_TOTAL_CLTV_EXPIRY_DELTA`].
+  uint32 max_total_cltv_expiry_delta = 2;
+
+  // The maximum number of paths that may be used by (MPP) payments.
+  // Defaults to [`DEFAULT_MAX_PATH_COUNT`].
```
benthecarman: The `defaults to` text here won't really work because it'll put the constant but not actually render/fetch the value. Will need to just put the actual value here sadly

ldk-server/src/api/bolt11_send.rs
```diff
@@ -12,10 +13,17 @@ pub(crate) fn handle_bolt11_send_request(
 	let invoice = Bolt11Invoice::from_str(&request.invoice.as_str())
 		.map_err(|_| ldk_node::NodeError::InvalidInvoice)?;
 
+	let route_parameters = request.route_parameters.map(|params| RouteParametersConfig {
+		max_total_routing_fee_msat: params.max_total_routing_fee_msat,
+		max_total_cltv_expiry_delta: params.max_total_cltv_expiry_delta,
+		max_path_count: params.max_path_count as u8,
+		max_channel_saturation_power_of_half: params.max_channel_saturation_power_of_half as u8,
```
benthecarman: since we are going down from u32 to u8, this isn't really safe. we should use `try_into` here so we can properly handle overflows

Anyitechs: Done

PR #80, opened 2025-12-09
Limit request body size to 10MB

benthecarman:
Limit the request body to prevent abuse and too much memory allocation


tnull (approved):
Makes sense!

PR #81, opened 2025-12-10
Hook up ChannelConfig in open channel endpoint & add to cli

benthecarman:
We had this in the proto files but was not hooked up on the server side or on the cli. This adds it on both ends

### DIFF
--- a/ldk-server-cli/src/main.rs
+++ b/ldk-server-cli/src/main.rs
@@ -11,7 +11,8 @@ use ldk_server_client::ldk_server_protos::api::{
 	OpenChannelRequest,
 };
 use ldk_server_client::ldk_server_protos::types::{
-	bolt11_invoice_description, Bolt11InvoiceDescription, PageToken, Payment,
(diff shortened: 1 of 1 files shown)


benthecarman:
I saw that todo but it seemed separate. That is for updating the channel config, not setting it on open

ldk-server-cli/src/main.rs
```diff
@@ -101,6 +102,29 @@ enum Commands {
...
+			help = "Max dust HTLC exposure as fixed limit (in milli-satoshi)",
+			conflicts_with = "max_dust_htlc_fee_rate_multiplier"
+		)]
+		max_dust_htlc_fixed_limit_msat: Option<u64>,
+		#[arg(
+			long,
+			help = "Max dust HTLC exposure as fee rate multiplier",
+			conflicts_with = "max_dust_htlc_fixed_limit_msat"
+		)]
+		max_dust_htlc_fee_rate_multiplier: Option<u64>,
```
wpaulino: Some of these options could really use some more documentation, but at the same time, I wouldn't expect users to set these in most cases. Maybe we just keep the forwarding specific ones for now and revisit the others if there's interest from users?

PR #82, opened 2025-12-10
Move endpoint path constants to shared defintion

benthecarman:
We had these defined in 2 separate areas which defeats the point of having them defined in constants. This moves them into the ldk-server-protos crate so they are shared among both the cli and server. Luckily there weren't any discrepancies.


tnull (approved):
Simple DRY up. Makes sense.

PR #84, opened 2025-12-11
Implement proper pagination for list-payments cli

benthecarman:
Another thing in the backend but not the cli

### DIFF
--- a/ldk-server-cli/src/main.rs
+++ b/ldk-server-cli/src/main.rs
@@ -24,10 +24,13 @@ use ldk_server_client::ldk_server_protos::api::{
 	SpliceOutResponse, UpdateChannelConfigRequest, UpdateChannelConfigResponse,
 };
 use ldk_server_client::ldk_server_protos::types::{
-	bolt11_invoice_description, Bolt11InvoiceDescription, ChannelConfig, PageToken, Payment,
+	bolt11_invoice_description, Bolt11InvoiceDescription, ChannelConfig, PageToken,
 	RouteParametersConfig,
 };
 use serde::Serialize;
+use types::CliListPaymentsResponse;
+
+mod types;
 
 // Having these default values as constants in the Proto file and
 // importing/reusing them here might be better, but Proto3 removed
@@ -178,9 +181,12 @@ enum Commands {
 	ListPayments {
 		#[arg(short, long)]
 		#[arg(
-			help = "Minimum number of payments to return. If not provided, only the first page of the paginated list is returned."
+			help = "Fetch at least this many payments by iterating through multiple pages. Returns combined results with the last page token. If not provided, returns only a single page."
 		)]
 		number_of_payments: Option<u64>,
+		#[arg(long)]
+		#[arg(help = "Page token to continue from a previous page (format: token:index)")]
+		page_token: Option<String>,
 	},
 	UpdateChannelConfig {
 		#[arg(short, long)]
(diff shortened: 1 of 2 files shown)


ldk-server-cli/src/main.rs
```diff
@@ -106,9 +106,15 @@ enum Commands {
...
-			help = "Minimum number of payments to return. If not provided, only the first page of the paginated list is returned."
+			help = "Fetch at least this many payments by iterating through multiple pages. Returns combined results with the last page token. If not provided, returns only a single page."
 		)]
 		number_of_payments: Option<u64>,
+		#[arg(long)]
+		#[arg(help = "Page token value to continue from a previous page")]
+		page_token_value: Option<String>,
+		#[arg(long)]
+		#[arg(help = "Page token index to continue from a previous page")]
+		page_token_index: Option<i64>,
```
tankyleo: Any chance we can make these into a single argument to eliminate footguns ? As long as we can display the `PageToken` as a single object that they can directly copy into the single argument here.

ldk-server-cli/src/main.rs
```diff
@@ -230,30 +236,64 @@ async fn main() {
...
+async fn handle_list_payments(
+	client: LdkServerClient, number_of_payments: Option<u64>, page_token_value: Option<String>,
+	page_token_index: Option<i64>,
+) -> Result<ListPaymentsResponse, LdkServerError> {
+	if page_token_value.is_some() != page_token_index.is_some() {
+		return Err(LdkServerError::new(
+			InternalError,
+			"Both --page-token-value and --page-token-index must be provided together".to_string(),
+		));
+	}
```
tankyleo: Single argument eliminates the possibility of this error here

ldk-server-cli/src/main.rs
```diff
@@ -230,30 +236,64 @@ async fn main() {
...
+	let mut next_page_token;
 
 	loop {
 		let response = client.list_payments(ListPaymentsRequest { page_token }).await?;
 
 		payments.extend(response.payments);
-		if payments.len() >= target_count as usize || response.next_page_token.is_none() {
+		next_page_token = response.next_page_token;
+
+		if payments.len() >= target_count as usize || next_page_token.is_none() {
```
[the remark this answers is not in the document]
benthecarman: Yes

ldk-server-cli/src/main.rs
benthecarman: done!

ldk-server-cli/src/types.rs
```diff
@@ -0,0 +1,41 @@
...
+	fn from(response: ListPaymentsResponse) -> Self {
+		let next_page_token = response.next_page_token.map(format_page_token);
+
+		CliListPaymentsResponse { payments: response.payments, next_page_token }
+	}
+}
+
+fn format_page_token(token: PageToken) -> String {
+	format!("{}:{}", token.token, token.index)
+}
```
tankyleo: Is this a better fit in the `fn handle_list_payments_request` in `ldk-server` ? Pretty much delete the `PageToken` type in the proto definition, and switch it to `optional string next_page_token = 2`.

Like this people who interact with the server programmatically also no longer can mishandle these two pieces ?

Although I agree they get a single object of type `PageToken`, so may be good enough there.

tankyleo: should we be concerned with breaking things at the moment ? I'd say we should be breaking liberally until a first release :)

tankyleo: discussed offline, for now we prefer formatting for CLI output in `ldk-server-cli`

PR #86, opened 2025-12-17
Fix clippy issues, enforce in CI

benthecarman:
We weren't checking for any clippy issues. This fixes most of them and adds a few exceptions for ones not worth fixing or ones that will be fixed with future changes.

### DIFF
--- a/ldk-server/src/api/error.rs
+++ b/ldk-server/src/api/error.rs
@@ -38,12 +38,10 @@ impl fmt::Display for LdkServerError {
 }
 
 #[derive(Clone, Debug, PartialEq, Eq)]
+#[allow(clippy::enum_variant_names)]
 pub(crate) enum LdkServerErrorCode {
 	/// Please refer to [`protos::error::ErrorCode::InvalidRequestError`].
 	InvalidRequestError,
-
-	/// Please refer to [`protos::error::ErrorCode::AuthError`].
-	AuthError,
 
 	/// Please refer to [`protos::error::ErrorCode::LightningError`].
 	LightningError,
@@ -56,7 +54,6 @@ impl fmt::Display for LdkServerErrorCode {
 	fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
 		match self {
 			LdkServerErrorCode::InvalidRequestError => write!(f, "InvalidRequestError"),
-			LdkServerErrorCode::AuthError => write!(f, "AuthError"),
 			LdkServerErrorCode::LightningError => write!(f, "LightningError"),
 			LdkServerErrorCode::InternalServerError => write!(f, "InternalServerError"),
 		}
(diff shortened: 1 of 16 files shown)


ldk-server/src/io/events/event_publisher.rs
```diff
@@ -41,6 +41,8 @@ pub trait EventPublisher: Send + Sync {
 	async fn publish(&self, event: EventEnvelope) -> Result<(), LdkServerError>;
 }
 
+/// A no-op implementation of the [`EventPublisher`] trait.
+#[allow(dead_code)] // This is unused when rabbitmq feature is enabled
```
tnull: Why not cfg-gate it on `not(feature = "events-rabbitmq")` then?

ldk-server/src/io/persist/sqlite_store/mod.rs
tnull: Should be okay for now, but that's a reminder that before we have the codebases diverge even further, can should finally upstream `PaginatedKVStore` to LDK and add SqliteStore/PostgresStore there, too.

ldk-server/src/util/proto_adapter.rs
```diff
@@ -378,6 +374,7 @@ pub(crate) fn pending_sweep_balance_to_proto(
 	}
 }
 
+#[allow(clippy::too_many_arguments)]
```
Anyitechs: I believe this is one of the exceptions being mentioned in the commit message as refactoring the function here is more suitable than suppressing the warning. Could you add a TODO here?

ldk-server/src/api/error.rs
```diff
@@ -29,11 +29,13 @@ impl fmt::Display for LdkServerError {
 }
 
 #[derive(Clone, Debug, PartialEq, Eq)]
+#[allow(clippy::enum_variant_names)]
 pub(crate) enum LdkServerErrorCode {
 	/// Please refer to [`protos::error::ErrorCode::InvalidRequestError`].
 	InvalidRequestError,
 
 	/// Please refer to [`protos::error::ErrorCode::AuthError`].
+	#[allow(unused)] // todo implement auth
```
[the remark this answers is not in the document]
benthecarman: my plan was to do this next, so it'll be used soon

ldk-server/src/io/persist/mod.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(dead_code)]
```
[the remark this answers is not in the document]
tnull: I don't know? Seems the code is a mess here with half-finished and half-migrated things all over the place. Seems we currently don't even use `PaginatedKVStore::remove` anywhere. I think rather than just hide the warnings we should rip out anything that is not used as we might or might not use it in the future, and if we'd do we need to revisit everything line-by-line either way as especially (async) IO code might be error prone anyways.

I say let's drop anything that gives an unused warning, add some crate-wide `deny` statements (the usual dead code, broken docs, etc) and then go from there.

PR #88, opened 2025-12-18
Add HMAC-based authentication for RPC/CLI

benthecarman:
Implements time-based HMAC-SHA256 authentication using a shared API key.
Each request includes a timestamp and HMAC in the X-Auth header, preventing
replay attacks with a 60-second tolerance window.

### DIFF
--- a/ldk-server-client/src/client.rs
+++ b/ldk-server-client/src/client.rs
@@ -13,6 +13,8 @@ use crate::error::LdkServerError;
 use crate::error::LdkServerErrorCode::{
 	AuthError, InternalError, InternalServerError, InvalidRequestError, LightningError,
 };
+use bitcoin_hashes::hmac::{Hmac, HmacEngine};
+use bitcoin_hashes::{sha256, Hash, HashEngine};
 use ldk_server_protos::api::{
 	Bolt11ReceiveRequest, Bolt11ReceiveResponse, Bolt11SendRequest, Bolt11SendResponse,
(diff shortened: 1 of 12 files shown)


ldk-server-client/src/client.rs
```diff
@@ -196,10 +198,12 @@ impl LdkServerClient {
 		&self, request: &Rq, url: &str,
 	) -> Result<Rs, LdkServerError> {
 		let request_body = request.encode_to_vec();
+		let (username, password) = &self.auth_credentials;
 		let response_raw = self
 			.client
 			.post(url)
 			.header(CONTENT_TYPE, APPLICATION_OCTET_STREAM)
+			.basic_auth(username, Some(password))
```
tnull: I don't think we should *ever* transmit an unhashed and unsalted password.

tnull (commented):
So far we assumed the RPC API to not be publicly accessible. If we do assume it is, we should probably also start to take futher precautions (e.g., take a look at DoS protection, etc).

That said, I'm not against adding authentication to the RPC protocol, however, if we do:

1) We should never transmit unhashed & unsalted passwords/credentials.
2) Authentication's utility is very limited if we're sending credentials over unencrypted channels (actually, it might just give a false sense of security). So if we add authentication, we should probably start looking into (requiring) TLS for the RPC connections, and add corresponding helpers to generate and configure corresponding self-signed certificates.

benthecarman:
Used claude to swith to a HMAC based approach instead, this way we aren't leaking the secrets in the request, working on the TLS stuff in a follow up

tnull (approved):

PR #92, opened 2026-01-09
Add TLS support for RPC/CLI 

benthecarman:
Built on top of #88 to reduce conflicts

With TLS we can now have encrypted communication with the client and
server. We auto generate a self signed cert on startup.

Co-Authored-By: Claude Opus 4.5 <noreply@anthropic.com>

### DIFF
--- a/ldk-server/src/util/tls.rs
+++ b/ldk-server/src/util/tls.rs
@@ -1,0 +1,226 @@
+// This file is Copyright its original authors, visible in version control
+// history.
+//
+// This file is licensed under the Apache License, Version 2.0 <LICENSE-APACHE
+// or http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your option.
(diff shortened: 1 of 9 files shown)


ldk-server/src/util/tls.rs
```diff
@@ -0,0 +1,227 @@
...
+const PEM_KEY_END: &str = "-----END PRIVATE KEY-----";
+
+/// Gets or generates TLS configuration. If custom paths are provided, uses those.
+/// Otherwise, generates a self-signed certificate in the storage directory.
+pub fn get_or_generate_tls_config(
+	tls_config: Option<TlsConfig>, storage_dir: &str,
+) -> Result<ServerConfig, String> {
+	if let Some(tls) = tls_config {
+		// Custom hosts but auto-generated cert
+		let cert_path = tls.cert_path.unwrap_or(format!("{storage_dir}/tls_cert.pem"));
```
tnull: I think `.crt`/`.key` suffixes are pretty common, especially the key one. Maybe worth renaming?

tnull (approved):

PR #100, opened 2026-01-18
Add announcement address setting

tankyleo:


benthecarman (changes requested):
Can we make this setting allow for multiple set? Since ldk takes in multiple, may as well do the same here

PR #101, opened 2026-01-19
Squash some typos

tankyleo:


ldk-server/src/util/config.rs
```diff
@@ -261,7 +261,7 @@ struct EsploraConfig {
 #[derive(Deserialize, Serialize)]
 struct LogConfig {
 	level: Option<String>,
-	file: Option<String>,
```
benthecarman: why update this and not the example file?

PR #102, opened 2026-01-19
Add help text for all cli commands and args

benthecarman:
For most of these I tried to just copy from the proto docs or ldk-node docs, adapting the wording when necessary.


tnull (approved):
LGTM. 

Would be cool to eventually automate this, maybe through some kind of template so we avoid the docs becoming stale.

PR #107, opened 2026-01-21
Add ConnectPeer API endpoint

joostjager:
Adds a new endpoint to connect to a peer on the Lightning Network without opening a channel. This is useful for establishing connections before channel operations or for maintaining peer connectivity.

The endpoint accepts node_pubkey, address, and an optional persist flag that defaults to true for automatic reconnection on restart.

Needed this endpoint when working on #108. Nodes sometimes seems very slow to auto-reconnect, or not reconnect at all?


benthecarman (commented):
can you add to the cli as well?

PR #111, opened 2026-01-22
Add LDK Server GUI

toneloc:
This adds an all-Rust [egui](https://github.com/emilk/egui) front-end for LDK Server.

This can visually showcase all of the novel functionality that LDK has recently added, like splicing and LSP/liquidity capabilities, while keeping the project all in Rust and maintaining deployment flexibility.

Perhaps it will be useful for devs and testers to poke around and to get real-world payments testing on the project.

Egui for me has worked well for me across Linux, Windows, and Mac desktops. With a few changes, this likely can compile into WASM for web use.

What do you think?

Screenshot 2026-01-22 at 10 50 21 AM

Screenshot 2026-01-22 at 11 35 00 AM

Screenshot 2026-01-22 at 10 51 16 AM

Screenshot 2026-01-22 at 10 50 31 AM

Screenshot 2026-01-22 at 11 09 59 AM


toneloc:
Makes total sense. I want this anyway for my LSP, so I'll continue to maintain in parallel as you guys make updates here. 

I know this repo is fast-moving at the moment, so I'll revert when I have something stabler to show. 

Thanks for the nice offer and kudos to the team for shipping this much-needed repo.

PR #117, opened 2026-01-27
Introduce telemetry for observability

Anyitechs:
This introduces the foundational telemetry infrastructure to improve
the observability of LDK Server.

It adds a new `/metrics` endpoint exposed on the REST service address,
which serves Prometheus-compatible metrics. This endpoint is public and
does not require HMAC authentication, allowing for easy integration with
monitoring systems.

- Added a `Metrics` utility struct to hold all the metrics we need to
expose.
- Made `Metrics` configurable via config.
- Added Prometheus basic auth support.

This is the first step in a larger effort to provide comprehensive telemetry.
Future updates will expand this to include other detailed metrics for channels,
balances, payments, etc.

Related issue #38

### DIFF
--- a/e2e-tests/tests/e2e.rs
+++ b/e2e-tests/tests/e2e.rs
@@ -12,7 +12,8 @@ use std::time::Duration;
 
 use e2e_tests::{
 	find_available_port, mine_and_sync, run_cli, run_cli_raw, setup_funded_channel,
-	wait_for_onchain_balance, LdkServerHandle, RabbitMqEventConsumer, TestBitcoind,
+	wait_for_onchain_balance, LdkServerConfig, LdkServerHandle, RabbitMqEventConsumer,
+	TestBitcoind,
 };
 use hex_conservative::{DisplayHex, FromHex};
 use ldk_node::bitcoin::hashes::{sha256, Hash};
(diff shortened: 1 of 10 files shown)


ldk-server/src/util/metrics.rs
```diff
@@ -0,0 +1,172 @@
...
+
+use std::sync::atomic::{AtomicI64, Ordering};
+use std::time::Duration;
+
+use ldk_node::Node;
+
+pub const BUILD_METRICS_INTERVAL: Duration = Duration::from_secs(60);
+
+/// This represents a [`Metrics`] type that can go up and down in value.
+pub struct IntGauge {
```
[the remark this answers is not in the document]
Anyitechs: A plain `AtomicI64` could work, but I introduce the type for better organization and to represent the `Gauge` metric type (this will help differentiate it from a `Counter` type when introduced later, which could also use a `AtomicI64` type but are meant to only increase and not decrease, in the metrics world). 

Happy to drop if it's too much boilerplate.

ldk-server/src/util/metrics.rs
```diff
@@ -0,0 +1,172 @@
...
+			"gauge",
+			&self.service_health_score.get().to_string(),
+		);
+
+		self.format_metrics_output(&mut buffer, options);
+
+		buffer
+	}
+
+	fn compute_health_score(is_running: bool, has_peers: bool, is_wallet_synced: bool) -> i64 {
```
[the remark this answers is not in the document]
benthecarman: Lnd has a prometheus integration. Would be good to look at and see what they are exposing in theirs

ldk-server/src/service.rs
```diff
@@ -154,6 +157,17 @@ impl Service<Request<Incoming>> for NodeService {
...
 	fn call(&self, req: Request<Incoming>) -> Self::Future {
+		// Handle metrics endpoint separately to bypass auth and return plain text
+		if req.uri().path().len() > 1 && &req.uri().path()[1..] == GET_METRICS_PATH {
+			let metrics = Arc::clone(&self.metrics);
+			return Box::pin(async move {
+				Ok(Response::builder()
+					.header("Content-Type", "text/plain")
+					.body(Full::new(Bytes::from(metrics.gather_metrics())))
+					.unwrap())
+			});
```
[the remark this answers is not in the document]
Anyitechs: > We aren't validating auth here we are short cutting before its done.

Yes, this is intentional because Prometheus does not support the HMAC auth scheme we use. It supports only basic auth and TLS.

ldk-server/src/service.rs
```diff
@@ -159,6 +163,17 @@ impl Service<Request<Incoming>> for NodeService {
 	type Future = Pin<Box<dyn Future<Output = Result<Self::Response, Self::Error>> + Send>>;
 
 	fn call(&self, req: Request<Incoming>) -> Self::Future {
+		// Handle metrics endpoint separately to bypass auth and return plain text
+		if req.uri().path().len() > 1 && &req.uri().path()[1..] == GET_METRICS_PATH {
```
benthecarman: should also require the request is a GET request

ldk-server-client/src/client.rs
```diff
@@ -106,6 +113,18 @@ impl LdkServerClient {
 		self.post_request(&request, &url).await
 	}
 
+	/// Retrieve the node metrics in Prometheus format.
+	pub async fn get_metrics(&self) -> Result<String, LdkServerError> {
```
benthecarman: This is just a raw string, really should be decoded into the Response type

ldk-server/src/main.rs
```diff
@@ -256,6 +257,25 @@ fn main() {
 			}
 		};
 		let event_node = Arc::clone(&node);
+
+		let metrics_node = Arc::clone(&node);
+		let mut interval = tokio::time::interval(BUILD_METRICS_INTERVAL);
+		let metrics = Arc::new(Metrics::new());
```
benthecarman: We're still building and tracking these metrics when they aren't enabled, this adds some extra overhead that we don't need. Can we only do it when metrics are enabled?

e2e-tests/tests/e2e.rs
```diff
@@ -634,3 +634,63 @@ async fn test_forwarded_payment_event() {
...
+	assert!(metrics.contains("ldk_server_total_public_channels_count 0"));
+	assert!(metrics.contains("ldk_server_total_private_channels_count 0"));
+	assert!(metrics.contains("ldk_server_total_onchain_balance_sats 0"));
+	assert!(metrics.contains("ldk_server_spendable_onchain_balance_sats 0"));
+	assert!(metrics.contains("ldk_server_total_anchor_channels_reserve_sats 0"));
+	assert!(metrics.contains("ldk_server_total_lightning_balance_sats 0"));
+
+	// Set up channel and make a payment to trigger metric update
+	setup_funded_channel(&bitcoind, &server_a, &server_b, 100_000).await;
+
```
benthecarman: We should be able to check that the channel (private/public) count, our LN balance, and num peers, goes up after opening a channel. It also sends funds to the nodes for anchor channels so we should be able to verify our on-chain balance numbers change.

e2e-tests/tests/e2e.rs
```diff
@@ -778,3 +778,88 @@ async fn test_hodl_invoice_fail() {
...
+	// We need to poll here because the metrics below are poll-based and
+	// updates every 60 seconds.
+	let timeout = Duration::from_secs(65);
```
benthecarman: We can't really do this. Are these not updated from the events? If we still need this even with the events updating. We should just make the interval configurable and make it like every 1 sec for the tests.

ldk-server/src/main.rs
```diff
@@ -331,6 +353,10 @@ fn main() {
 								&event_node,
 								Arc::clone(&event_publisher),
 								Arc::clone(&paginated_store)).await;
+
+							if let Some(metrics) = &metrics {
+								metrics.update_payments_count(true);
```
benthecarman: we should be able to update our balance metric here and the payment received one

PR #124, opened 2026-02-14
Do not error on startup for optional values

elnosh:
If the config file didn't have `listening_addresses`, `announcement_addresses` or `dir_path` it would err and fail to start up. 

- For `dir_path` it would fail before it could fallback to the `~/.ldk-server/` default.  
- For `listening_addresses` and `announcement_addresses` I think those should be options as well (?) so removed those errors

### DIFF
--- a/ldk-server/src/util/config.rs
+++ b/ldk-server/src/util/config.rs
@@ -185,36 +185,39 @@ impl ConfigBuilder {
 			.parse::<SocketAddr>()
 			.map_err(|e| io::Error::new(io::ErrorKind::InvalidInput, e))?;
 
-		let storage_dir_path =
-			self.storage_dir_path.ok_or_else(|| missing_field_err("storage_dir_path"))?;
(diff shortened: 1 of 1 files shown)


ldk-server/src/util/config.rs
```diff
@@ -891,22 +901,55 @@ mod tests {
...
+	fn test_config_optional_values() {
+		let storage_path = std::env::temp_dir();
+		let config_file_name = "test_only_required_config.toml";
+
+		let mut args_config = empty_args_config();
+		args_config.config_file =
+			Some(storage_path.join(config_file_name).to_string_lossy().to_string());
+
+		// Test with optional values not specified in the config file
+		#[cfg(not(any()))]
```
benthecarman: why this?

elnosh: woops, residue from my debugging this test. Removed.

PR #125, opened 2026-02-16
Skip default config path in tests

benthecarman:
Tests calling `load_config` without an explicit `config_file` would fall back to `~/.ldk-server/config.toml` if it existed locally, causing test failures.

### DIFF
--- a/ldk-server/src/util/config.rs
+++ b/ldk-server/src/util/config.rs
@@ -20,12 +20,19 @@ use ldk_node::liquidity::LSPS2ServiceConfig;
 use log::LevelFilter;
 use serde::{Deserialize, Serialize};
 
-use crate::get_default_data_dir;
-
+#[cfg(not(test))]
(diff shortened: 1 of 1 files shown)


ldk-server/src/util/config.rs
```diff
@@ -543,7 +543,15 @@ pub fn load_config(args: &ArgsConfig) -> io::Result<Config> {
 	let config_file = if let Some(path) = &args.config_file {
 		Some(PathBuf::from(path))
 	} else {
-		get_default_config_path().filter(|path| path.exists())
+		// Skip the default config path during tests to avoid picking up a real ~/.ldk-server/config.toml locally
```
tnull: Makes sense, just wondering if this/a similar override should happen in `get_default_config_path` itself to ensure that it always applies in tests, independent of where it's called?

benthecarman: This was the only call site but yeah that'll be safer in the long run

tnull (approved):

PR #126, opened 2026-02-16
Make counterparty_node_id optional for close/force-close/update-channel-config

benthecarman:
When `counterparty_node_id` is not provided, the server resolves it by looking up the channel via user_channel_id from the channel list.


tnull:
> Pre-open two channels can have the same channel id so it's definitely required. Obviously the user id should be unique.

Well, note that the API here uses the `user_channel_id`, not the `channel_id`. Given we assert the former is unique, we could theoretically drop the `counterparty_node_id` from the interface.

PR #128, opened 2026-02-19
Add end-to-end test framework

### LINKED ISSUE #113: End-to-end testing for ldk-server
Currently, ldk-server lacks end-to-end testing coverage for the cli→client→server pathway. This makes it difficult to validate that changes work correctly across all layers without manual testing.

See also https://github.com/lightningdevkit/ldk-server/pull/109#issuecomment-3784513457

benthecarman:
Closes #113 

Introduces an e2e-tests crate that spins up real bitcoind and ldk-server processes to test the cli -> ldk-server flow. A build.rs automatically compiles ldk-server (with events-rabbitmq) and ldk-server-cli into a separate target directory, so tests work without manual pre-building.

Tests cover all cli commands and most rabbitmq events.

Claude helped with some of the boilerplate code

Found a small place we can make the bolt12 cli better

### DIFF
--- a/.github/workflows/integration-tests-events-rabbitmq.yml
+++ b/.github/workflows/integration-tests-events-rabbitmq.yml
@@ -28,6 +28,23 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v4
 
+      - name: Enable caching for bitcoind
+        id: cache-bitcoind
+        uses: actions/cache@v4
+        with:
+          path: bin/bitcoind-${{ runner.os }}-${{ runner.arch }}
+          key: bitcoind-${{ runner.os }}-${{ runner.arch }}
+
+      - name: Download bitcoind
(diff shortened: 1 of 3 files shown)


ldk-server-cli/src/main.rs
```diff
@@ -158,12 +158,12 @@ enum Commands {
 	},
 	#[command(about = "Return a BOLT12 offer for receiving payments")]
 	Bolt12Receive {
+		#[arg(help = "Description to attach along with the offer")]
```
tnull: Why was this moved?

benthecarman: Something I noticed making the bolt12 test, we can't create an amountless bolt12 because the description is required but the amount is optional. I'll move to a separate commit to make it cleaner

benthecarman:
working on more tests for here, but can always do as a follow up if we want to merge now

tnull (approved):

PR #136, opened 2026-02-25
Allow hostname RPC support

### LINKED ISSUE #66: Allow hostname-based RPC communication (remove SocketAddr restriction)
### Problem
Currently, LDK Server expects the Bitcoin RPC address to be passed as a `host:port` string, which it then tries to parse using Rust’s SocketAddr. However, SocketAddr only supports IP addresses (IPv4/IPv6) — hostnames are not allowed.

This is a problem in containerized environments (like **Docker**), where inter-service communication is usually done via hostnames (e.g., `bitcoind:18443`). Since **SocketAddr** rejects hostnames, users are forced to use hardcoded IPs, which can change if a container restarts, making the setup fragile.

**LDK Server** also extracts the IP and port from **SocketAddr** to pass them separately into `set_chain_source_bitcoind_rpc`, so even if we could parse the string ourselves, this separation would still be necessary.

### Example use case

Running ldk-server and bitcoind in separate containers on the same Docker network. Ideally, I want to set:

``` toml
rpc_address = "bitcoind:18443"
```
But this fails because "bitcoind" is not a valid IP address.

### Proposed solution

A clean solution — already used in other projects like Eclair and LND — is to split the configuration into two fields:

```toml
rpc_host = "bitcoind"
rpc_port = 18443
```
This avoids the limitations of **SocketAddr** entirely and fits better with containerized setups.

Considerations

This would require changing how validation is performed, since the current implementation relies on SocketAddr parsing.

Anyitechs:
This removes the restriction of `SocketAddr` and allows us to support hostnames, which is essential in containerized/dynamic environments.

Closes #66

### DIFF
--- a/ldk-server/src/main.rs
+++ b/ldk-server/src/main.rs
@@ -152,13 +152,8 @@ fn main() {
 	}
 
 	match config_file.chain_source {
-		ChainSource::Rpc { rpc_address, rpc_user, rpc_password } => {
-			builder.set_chain_source_bitcoind_rpc(
-				rpc_address.ip().to_string(),
-				rpc_address.port(),
-				rpc_user,
-				rpc_password,
-			);
+		ChainSource::Rpc { rpc_host, rpc_port, rpc_user, rpc_password } => {
+			builder.set_chain_source_bitcoind_rpc(rpc_host, rpc_port, rpc_user, rpc_password);
 		},
 		ChainSource::Electrum { server_url } => {
 			builder.set_chain_source_electrum(server_url, None);
--- a/ldk-server/src/util/config.rs
+++ b/ldk-server/src/util/config.rs
@@ -62,7 +62,7 @@ pub struct TlsConfig {
 
 #[derive(Debug, PartialEq, Eq)]
 pub enum ChainSource {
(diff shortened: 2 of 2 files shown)


ldk-server/src/main.rs
```diff
@@ -152,13 +152,8 @@ fn main() {
...
 	match config_file.chain_source {
-		ChainSource::Rpc { rpc_address, rpc_user, rpc_password } => {
-			builder.set_chain_source_bitcoind_rpc(
-				rpc_address.ip().to_string(),
-				rpc_address.port(),
-				rpc_user,
-				rpc_password,
-			);
+		ChainSource::Rpc { rpc_host, rpc_port, rpc_user, rpc_password } => {
+			builder.set_chain_source_bitcoind_rpc(rpc_host, rpc_port, rpc_user, rpc_password);
```
tankyleo: Up to you people to make the call, I'd prefer parsing a single `<hostname>:<port>` string rather than splitting that setting into two parameters. This is what I do in VSS, it optimizes for fewer config params. That setting is logically a single thing right?

benthecarman (approved):

PR #142, opened 2026-03-03
Add e2e config startup tests

benthecarman:
Continuation of #128 / #113

Add config test suite that verifies server startup with various config combinations (to prevent things like #129) and validates errors for invalid configs.

### DIFF
--- a/e2e-tests/src/lib.rs
+++ b/e2e-tests/src/lib.rs
@@ -354,6 +354,46 @@ fn spawn_server(
 	(child, params)
 }
 
+/// Start ldk-server with the given config and expect it to fail (exit non-zero).
+/// Returns the stderr output for assertion in tests.
+pub fn start_expect_failure(
+	bitcoind: &TestBitcoind, config_fn: impl FnOnce(&TestServerParams) -> String,
+) -> String {
+	let (mut child, ..) = spawn_server(bitcoind, config_fn);
+
+	let timeout = Duration::from_secs(30);
+	let start = std::time::Instant::now();
+	loop {
+		match child.try_wait() {
+			Ok(Some(_)) => break,
+			Ok(None) => {
+				if start.elapsed() > timeout {
+					let _ = child.kill();
+					panic!(
+						"Server did not exit within {:?} — it may have started successfully \
+						 instead of failing",
+						timeout
+					);
+				}
+				std::thread::sleep(Duration::from_millis(100));
+			},
+			Err(e) => panic!("Failed to wait for ldk-server process: {}", e),
+		}
+	}
+
+	let output = child
+		.wait_with_output()
+		.unwrap_or_else(|e| panic!("Failed to read ldk-server output: {}", e));
+
+	assert!(
+		!output.status.success(),
+		"Expected server to fail but it exited with status: {}",
+		output.status
+	);
+
+	String::from_utf8_lossy(&output.stderr).to_string()
+}
+
 /// Find an available TCP port by binding to port 0.
 pub fn find_available_port() -> u16 {
 	let listener = TcpListener::bind("127.0.0.1:0").unwrap();
--- a/e2e-tests/src/lib.rs
+++ b/e2e-tests/src/lib.rs
@@ -84,35 +84,54 @@ impl TestBitcoind {
 	}
 }
 
-/// Handle to a running ldk-server child process.
-pub struct LdkServerHandle {
-	child: Option<Child>,
+/// Wrapper around an electrsd process providing both Electrum and Esplora endpoints.
+pub struct TestElectrs {
+	pub electrsd: electrsd::ElectrsD,
+}
+
+impl TestElectrs {
+	/// Start an electrs instance connected to the given bitcoind with Esplora HTTP enabled.
+	pub fn new(bitcoind: &TestBitcoind) -> Self {
+		let mut conf = electrsd::Conf::default();
+		conf.http_enabled = true;
+		let electrsd =
+			electrsd::ElectrsD::with_conf(electrsd::exe_path().unwrap(), &bitcoind.bitcoind, &conf)
+				.unwrap();
+		Self { electrsd }
+	}
+
+	pub fn electrum_url(&self) -> String {
+		// electrsd binds to 0.0.0.0 but that's not a connectable address for clients
+		self.electrsd.electrum_url.replace("0.0.0.0", "127.0.0.1")
+	}
+
+	pub fn esplora_url(&self) -> String {
+		let url = self.electrsd.esplora_url.as_ref().expect("esplora not enabled");
+		// electrsd binds to 0.0.0.0 but that's not a connectable address for clients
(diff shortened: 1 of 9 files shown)


Anyitechs (approved):
Thanks for the cleanup in `config.rs`, and the robust test!

The logic looks great to me. Just needs a rebase to clear the conflicts.

Camillarhi:
Thanks for this. Building this branch with `cargo build` emits `dead_code` warnings on `rabbitmq_connection_string`, `rabbitmq_exchange_name`, and `lsps2_service_config` in `ldk-server/src/util/config.rs`. Looks like the `#[allow(dead_code)]` guard on those fields was dropped when they became `Option`.

benthecarman:
Cleaned up to be a lot simpler than previous iteration

e2e-tests/src/lib.rs
```diff
@@ -141,21 +157,81 @@ min_payment_size_msat = 0
...
+	client: LdkServerClient,
+	// Kept alive so the electrs process doesn't get dropped
+	_electrs: Option<TestElectrs>,
+}
 
-		let config_path = storage_dir.join("config.toml");
-		std::fs::write(&config_path, &config_content).unwrap();
+impl LdkServerHandle {
+	/// Starts a new ldk-server instance against the given bitcoind.
+	/// Randomly picks between bitcoind RPC, electrum, and esplora as the chain source.
```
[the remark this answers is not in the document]
joostjager: seems dropped from this pr now, so no discussion point anymore

joostjager (commented):
E2E tests seem to be pretty heavy for config validation. Can't that be done just as well on the unit test level?

Perhaps a smoke test is useful that tests one invalid config, see how that traverses the full stack.

benthecarman:
> E2E tests seem to be pretty heavy for config validation. Can't that be done just as well on the unit test level?
> 
> Perhaps a smoke test is useful that tests one invalid config, see how that traverses the full stack.

Yeah we already have config unit tests, the goal of these is to make it so we test things that our config can't really validate and also make sure we are properly handling a valid config. Removed some of the tests that were basically recreations of the config unit tests and added more that actually target where we could and should throw errors.

e2e-tests/tests/config.rs
```diff
@@ -0,0 +1,164 @@
...
+			"listening_addresses =",
+			&format!(
+				"listening_addresses = [\"127.0.0.1:{}\", \"127.0.0.1:{}\"]",
+				params.p2p_port, extra_port
+			),
+		)
+	})
+	.await;
+	let info = server.client().get_node_info(GetNodeInfoRequest {}).await.unwrap();
+	assert!(info.current_best_block.is_some());
```
joostjager, line 55: Here and above, perhaps assert a bit more, in particular the addresses?

These tests do look quite similar to what's already in ldk-node, IIRC. Not sure if it adds much.

e2e-tests/src/lib.rs
```diff
@@ -157,21 +160,45 @@ enabled = true
...
+	/// Starts a new ldk-server instance against the given bitcoind.
+	/// Waits until the server is ready to accept requests.
+	pub async fn start(bitcoind: &TestBitcoind) -> Self {
+		Self::start_with_options(bitcoind, LdkServerConfig::default()).await
+	}
+
+	pub async fn start_with_options(bitcoind: &TestBitcoind, config: LdkServerConfig) -> Self {
+		Self::start_with_config(bitcoind, |params| {
+			let chain_source = format!(
+				"[bitcoind]\nrpc_address = \"{}\"\nrpc_user = \"{}\"\nrpc_password = \"{}\"",
```
joostjager: This is duplicated in a few places, perhaps it can be a helper.

e2e-tests/tests/config.rs
```diff
@@ -0,0 +1,164 @@
...
+	default_test_config, start_expect_failure, test_config_with_chain_source, LdkServerHandle,
+	TestBitcoind,
+};
+use ldk_server_grpc::api::GetNodeInfoRequest;
+
+fn remove_config_line(config: &str, key: &str) -> String {
+	config.lines().filter(|line| !line.trim_start().starts_with(key)).collect::<Vec<_>>().join("\n")
+}
+
+fn replace_config_line(config: &str, key: &str, new_line: &str) -> String {
```
joostjager: I am not sure about this mix of having a config builder, but then also surgically modifying it again. Might be a bit brittle too. Any way to make this more consistent/robust?

joostjager (dismissed):
The latest shape is better after trimming config tests.

I do still get a slight feeling of overlap with ldk-node coverage. Going forward, I think we should be pretty strict about keeping that duplication to a minimum. Just test the things in ldk-server that also mainly live there.

Left a few non-blocking remarks.

e2e-tests/tests/config.rs
benthecarman: made this a lot cleaner

e2e-tests/src/lib.rs
benthecarman: fixed with the builder

e2e-tests/tests/config.rs
benthecarman: added more asserts

e2e-tests/src/lib.rs
```diff
@@ -103,42 +103,159 @@ pub struct LdkServerConfig {
...
+	pub rpc_address: String,
+	pub rpc_user: String,
+	pub rpc_password: String,
+}
+
+/// A chain source for the test config, mirroring the server's supported backends.
+pub enum ChainSource {
+	Bitcoind { rpc_address: String, rpc_user: String, rpc_password: String },
+	Electrum { server_url: String },
+	Esplora { server_url: String },
```
joostjager, line 120: Electrum and Esplora seem to be unused. Would remove it if it is dead currently.

e2e-tests/src/lib.rs
```diff
@@ -103,42 +103,159 @@ pub struct LdkServerConfig {
...
-		let (rpc_host, rpc_port_num, rpc_user, rpc_password) = bitcoind.rpc_details();
-		let rpc_address = format!("{rpc_host}:{rpc_port_num}");
+		let mut node = vec!["[node]".to_string(), "network = \"regtest\"".to_string()];
+		if !self.listening_addresses.is_empty() {
+			node.push(format!(
+				"listening_addresses = {}",
+				toml_string_array(&self.listening_addresses)
+			));
+		}
+		node.push(format!("grpc_service_address = \"{}\"", self.grpc_service_address));
```
joostjager, line 234: Could consider using `toml` which `ldk-server` already depends on. But this is fine too.

joostjager (approved):
Nicer indeed with the builder

e2e-tests/src/lib.rs
benthecarman: I am just gonna leave for now, we may implement it eventually.

PR #144, opened 2026-03-09
Add granular per-endpoint API key permissions

### LINKED ISSUE #140: Explore more granular auth schemes

benthecarman:
Closes #140 

Currently ldk-server's auth is binary, you either have full access or no access. This makes it difficult to give third-party applications (e.g. bots) access to a node without granting them the ability to send funds, force-close channels, or perform other sensitive operations.

This PR replaces the original single API key with a directory of multiple API keys with per-endpoint permissions. API keys are stored as toml files in <network_dir>/api_keys/, each specifying a hex key and allowed endpoints (or "*" for admin). Adds  `create-api-key` and `get-permissions` for creating api keys and to get your current auth permissions. Added some migration code to move your current api key to the new location. We can remove this eventually but also we could just do without it for now as we our the only real users so far.

To support multiple keys without iterating all of them on every request, the HMAC auth header is extended to include a key ID (HMAC `<key_id>:<timestamp>:<hmac>`), where key_id is the first 8 bytes of SHA256(api_key) hex-encoded. This allows O(1) key lookup on the server side.


benthecarman:
Opened #273 as alternative

PR #151, opened 2026-03-15
Add systemd readiness and stopping notifications

benthecarman:
If NOTIFY_SOCKET is set, send READY=1 and STOPPING=1 to systemd. This enables proper dependency ordering and faster startup detection when running as a Type=notify service unit.

Noticed this when deploying my own mainnet ldk-server on my server

### DIFF
--- a/ldk-server/src/util/systemd.rs
+++ b/ldk-server/src/util/systemd.rs
@@ -1,0 +1,68 @@
+// This file is Copyright its original authors, visible in version control
+// history.
+//
+// This file is licensed under the Apache License, Version 2.0 <LICENSE-APACHE
+// or http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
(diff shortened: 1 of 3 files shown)


ldk-server/src/util/systemd.rs
```diff
@@ -0,0 +1,64 @@
...
+		Err(e) => {
+			warn!("Failed to create socket for systemd notification: {e}");
+			return;
+		},
+	};
+
+	// systemd sets NOTIFY_SOCKET to either a filesystem path (e.g. /run/systemd/notify)
+	// or an abstract socket prefixed with '@'. Abstract sockets require special addressing.
+	let result = if let Some(abstract_name) = socket_path.strip_prefix('@') {
+		use std::os::linux::net::SocketAddrExt;
```
joostjager: Top of file uses?

joostjager (approved):

PR #154, opened 2026-03-17
Add Dockerfile

benthecarman:
Multi-stage build with dependency caching for faster rebuilds. Includes both ldk-server and ldk-server-cli binaries.


Dockerfile
```diff
@@ -0,0 +1,61 @@
...
+            -p ldk-server-cli; \
+    else \
+        cargo build --release --locked \
+            -p ldk-server \
+            -p ldk-server-cli; \
+    fi
+
+# Copy real source and rebuild
+COPY . .
+RUN touch ldk-server/src/main.rs ldk-server-cli/src/main.rs \
```
lorenzolfm, line 36: I think you're missing ldk-server-mcp here? You added the dummy for it above

benthecarman: No need for the mcp server in the docker build, that's something you run locally

Dockerfile
```diff
@@ -0,0 +1,61 @@
...
+# Copy manifests and lock file first for dependency caching
+COPY Cargo.toml Cargo.lock ./
+COPY ldk-server/Cargo.toml ldk-server/Cargo.toml
+COPY ldk-server-cli/Cargo.toml ldk-server-cli/Cargo.toml
+COPY ldk-server-client/Cargo.toml ldk-server-client/Cargo.toml
+COPY ldk-server-grpc/Cargo.toml ldk-server-grpc/Cargo.toml
+COPY ldk-server-grpc/build.rs ldk-server-grpc/build.rs
+COPY ldk-server-mcp/Cargo.toml ldk-server-mcp/Cargo.toml
+
+# Create dummy source files so cargo can resolve and build dependencies
```
lorenzolfm, line 15: you can use cargo-chef for this, btw. This way a workspace change would require a change in the dockerfile.

I understand wanting to avoid dependencies in the world we live in, tho haha

Dockerfile
```diff
@@ -0,0 +1,61 @@
...
+    && rm -rf /var/lib/apt/lists/*
+
+COPY --from=builder /app/target/release/ldk-server /usr/local/bin/ldk-server
+COPY --from=builder /app/target/release/ldk-server-cli /usr/local/bin/ldk-server-cli
+
+ENV LDK_SERVER_NODE_GRPC_SERVICE_ADDRESS=0.0.0.0:3536
+
+EXPOSE 9735 3536
+
+ENTRYPOINT ["ldk-server"]
```
lorenzolfm, line 61: wdyt about running this with a non-root user?

Dockerfile
benthecarman: Yeah this is easier to understand in my head too

Dockerfile
benthecarman: The node is pretty self contained, so the usual benefits aren't as much here. People can always run it as a different user in docker-compose if they want as well. Just want to the base docker image to be something simple to run

Dockerfile
lorenzolfm: Got it, maybe line 21 is irrelevant then

tankyleo (approved):
Passed this through multiple AIs

PR #168, opened 2026-03-24
Replace protobuf with JSON everywhere

benthecarman:
Protobuf added complexity without much benefit for our use case — the binary encoding is opaque, hard to debug with standard HTTP tools, and requires proto toolchain maintenance. JSON is human-readable, widely supported, and sufficient for our throughput needs.

This removes prost and all .proto files entirely, renaming the ldk-server-protos crate to ldk-server-json-models. Types are rewritten as hand-written Rust structs and enums with serde derives rather than prost-generated code. Fixed-size byte fields (hashes, channel IDs, public keys) use [u8; 32] and [u8; 33] with hex serde instead of String, giving type safety at the model layer.

Several proto-era patterns are cleaned up: wrapper structs that only existed because protobuf wraps oneof in a message are removed, fields that were Option only because proto message fields are nullable are made required where the server always provides them, and the EventEnvelope wrapper is dropped in favor of using Event directly.


ldk-server-json-models/src/types.rs
tnull: If we did this, I do wonder if we should also do https://github.com/lightningdevkit/ldk-node/issues/75 / continue https://github.com/lightningdevkit/rust-lightning/pull/3157 to have the serializations for LDK / LDK Node types live in their origin crate rather than replicating the types here? This would also avoid the types getting out-of-date, etc.

tnull (commented):
Thanks! This is def. a good base for more concrete discussion. Might make sense, although as mentioned elsewhere not loving JSON efficiency and its idiosyncrasies in general.

> Storage namespaces are changed from ("payments", "") to ("ldk-server", "payments") so existing protobuf-encoded data is silently ignored rather than failing to deserialize, avoiding the need for migration code or manual database wipes.

I don't think we should do this. Pre-0.1 we should be fine with breaking backwards compat rather than starting out with some  weird tech debt, especially since AFAIK it's only @tankyleo and myself that run Server in production currently.

tankyleo:
Thanks for the ping, good to do manual database operations.

ldk-server/src/main.rs
```diff
@@ -507,9 +504,9 @@ fn upsert_payment_details(
 	match paginated_store.write(
 		PAYMENTS_PERSISTENCE_PRIMARY_NAMESPACE,
 		PAYMENTS_PERSISTENCE_SECONDARY_NAMESPACE,
-		&payment.id,
+		&payment.id.to_lower_hex_string(),
 		time,
-		&payment.encode_to_vec(),
+		&serde_json::to_vec(&payment).unwrap(),
```
joostjager: Neither proto or json seems right here? I'd expect TLV to be used.

PR #176, opened 2026-03-27
Replace REST and RabbitMQ with gRPC

benthecarman:
Implement gRPC directly on hyper's HTTP/2 support, without tonic, to keep the dependency tree small. This consolidates two separate protocols (protobuf-over-REST for RPCs, RabbitMQ for events) into a single gRPC interface. 

#175 proved the migration works with tonic. We opted for from-scratch approach instead, informed by the plan at: https://gist.github.com/tnull/1f6ffc5ae71c418f28844e27eee8c62b

Events are now delivered via a server-streaming SubscribeEvents RPC backed by a tokio broadcast channel, eliminating the RabbitMQ operational dependency. HMAC auth is simplified to timestamp-only signing since TLS guarantees body integrity.

Had claude break it into several commits that should make it easy to review.

Also tested with the cli made in #175 and it works so seems our implementation is compliant


joostjager (commented):
Quick Claude check for missing parts

tnull (commented):
Started a first pass, so far reviewed the first 4 commits.

In general it will be hard to catch all the details, so the strategy likely is to drastically improve test coverage where we see the opportunity and go from there.

benthecarman:
fixed rebase conflicts

tnull:
> The reason we chose protobuf-over-rest earlier was due to web support, I don't see that being covered or discussed here? What are your thoughts on that? (There is grpc-web but it adds further complexity.)

Yes, this was the initial motivation for doing protobuf-over-REST for VSS Server, but not so much in LDK Server. Here we mostly followed the same approach, IIRC? 

Of course, protobuf would have always required a respective compiler/library in the web setting. So arguably, real/native web support would indeed mean JSON-over-REST, as @joostjager points out, which however has several idiosyncrasies of its own (starting with JSON being a super inefficient data format in general, but also stuff like only 56-bit integers, etc). And of course the reasons you've listed above on why protobuf might be preferable over JSON are valid (at least IMO). However, in turn, protobuf-over-REST was missing a few features we want / need (more integrated/ergonomic APIs for streaming notifications, etc pp).

Now, we concluded that either migrating to 'full' gRPC (to optimize for features and efficiency) or to JSON-over-REST (to optimize for compatibility with the web) would make sense. However, as discussed above both approaches also have their own cons, so we kind of expect that either way we'd end up with requiring additional software components that would allow us to supplement the weaknesses of either approach (depending on the use case). Thankfully, building and maintaining simple API surfaces that only delegate logic to core components is very very cheap since the recent improvements of AI tooling, so no big deal either way.

ldk-server/src/grpc.rs
```diff
@@ -23,8 +23,10 @@ use hyper::{Request, Response};
 use crate::api::error::{LdkServerError, LdkServerErrorCode};
 
 // gRPC status codes (a subset — only those we use).
+#[allow(dead_code)] // Used by streaming responses (SubscribeEvents)
```
tnull: Let's not even start with `dead_code` cfgs?

ldk-server/src/grpc.rs
```diff
@@ -0,0 +1,332 @@
...
+	}
+
+	Ok(())
+}
+
+/// Retrieve a gRPC metadata value from request headers.
+///
+/// Per the gRPC spec, headers ending in `-bin` contain base64-encoded binary data.
+/// All other metadata values are returned as raw UTF-8 bytes.
+#[allow(dead_code)] // Available for handlers that need to read binary gRPC metadata
```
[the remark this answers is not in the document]
tnull: Still unaddressed?

e2e-tests/tests/e2e.rs
```diff
@@ -842,27 +856,36 @@ async fn test_forwarded_payment_event() {
...
 
-	// Verify PaymentForwarded event on B
-	let events_b = consumer_b.consume_events(10, Duration::from_secs(15)).await;
-	assert!(
-		events_b.iter().any(|e| matches!(&e.event, Some(Event::PaymentForwarded(_)))),
-		"Expected PaymentForwarded event on LSP node B, got events: {:?}",
-		events_b.iter().map(|e| &e.event).collect::<Vec<_>>()
-	);
+	// Verify PaymentForwarded event on B (drain other events)
+	let forwarded = tokio::time::timeout(Duration::from_secs(30), async {
```
tnull, line 827: nit: Might be worth DRYing up the tooling here, and moving any timeouts to consts, etc?

ldk-server-client/src/client.rs
```diff
@@ -399,6 +401,54 @@ impl LdkServerClient {
...
 	}
 
+	/// Subscribe to a stream of server events via server-streaming gRPC.
+	///
+	/// Returns an [`EventStream`] that yields [`EventEnvelope`] messages as they arrive.
+	pub async fn subscribe_events(&self) -> Result<EventStream, LdkServerError> {
+		let request = SubscribeEventsRequest {};
+		let proto_bytes = request.encode_to_vec();
+		let mut grpc_body = Vec::with_capacity(5 + proto_bytes.len());
+		grpc_body.push(0u8);
```
tnull: Would even little stuff like this make it worth to move the gRPC out to a dedicated library in the workspace that both sides could reuse, maybe `ldk-server-protos` should become `ldk-server-grpc`? Or we introduce the latter alongside the former?

That might also the right place to add quite a bit of additional test coverage based on `tonic` and/or other pre-existing gRPC libraries to make sure we're compliant and interoperable?

ldk-server/src/api/bolt11_claim_for_hash.rs
```diff
@@ -15,9 +15,10 @@ use ldk_server_protos::api::{Bolt11ClaimForHashRequest, Bolt11ClaimForHashRespon
 use crate::api::error::LdkServerError;
 use crate::api::error::LdkServerErrorCode::InvalidRequestError;
 use crate::service::Context;
+use std::sync::Arc;
 
-pub(crate) fn handle_bolt11_claim_for_hash_request(
-	context: &Context, request: Bolt11ClaimForHashRequest,
+pub(crate) async fn handle_bolt11_claim_for_hash_request(
+	context: Arc<Context>, request: Bolt11ClaimForHashRequest,
```
tnull, line 21: Can we squash this commit into https://github.com/lightningdevkit/ldk-server/pull/176/changes/3c3a4558feb3c6ad89880aaec0033b81e44569a4 ?

tnull (commented):
Now did ~a full pass. Please squash the fixup commits into the feature commits first touching any of the given codepaths to at least make this PR a little less unwieldy.

Mostly looks good, I think, though one remaining question I have is whether we should give the gRPC stuff its own crate from the getgo.

ldk-server-client/src/client.rs
benthecarman: Renamed it to `ldk-server-grpc` and moved a lot of the grpc logic into there. Added some tests to test against tonic as well.

tnull (commented):
Needs a rebase now that #180 landed.

benthecarman:
rebased

ldk-server-client/src/client.rs
```diff
@@ -98,28 +89,26 @@ impl LdkServerClient {
...
-	fn compute_auth_header(&self, body: &[u8]) -> String {
+	/// Uses timestamp-only HMAC (no body) since TLS guarantees integrity.
+	fn compute_auth_header(&self) -> String {
 		let timestamp = SystemTime::now()
 			.duration_since(UNIX_EPOCH)
 			.expect("System time should be after Unix epoch")
 			.as_secs();
 
-		// Compute HMAC-SHA256(api_key, timestamp_bytes || body)
+		// HMAC-SHA256(api_key, timestamp_bytes) — no body
```
tnull, line 99: Remind me why we're fine with this now? And related test-changes landed in the previous commit, btw.

ldk-server-client/src/client.rs
```diff
@@ -144,411 +146,346 @@ impl LdkServerClient {
...
+		Rs::decode(proto_bytes).map_err(|e| {
+			LdkServerError::new(InternalError, format!("Failed to decode gRPC response: {}", e))
+		})
+	}
+}
+
+/// Map a gRPC status code to an LdkServerError.
+fn grpc_code_to_error(code: u32, message: String) -> LdkServerError {
+	let error_code = match code {
+		3 => InvalidRequestError,  // INVALID_ARGUMENT
```
tnull, line 458: Can we make these proper const in `ldk-server-grpc` rather than just comments?

ldk-server/src/service.rs
```diff
@@ -343,6 +360,52 @@ impl Service<Request<Incoming>> for NodeService {
...
+												.send(Err(GrpcStatus::new(
+													GRPC_STATUS_UNAVAILABLE,
+													"server shutting down",
+												)))
+												.await;
+											break;
+										},
+									}
+								},
+								_ = shutdown_rx.changed() => {
```
tnull, line 394: Should this be the first branch and we should be aborting as fast as possible via `biased;`?

tnull (approved):
Did another pass, but there's a lot going on. 

A few comments, but I'm still going ahead and landing this so it's out of the way and we can make more incremental changes without introducing rebase conflicts over and over again. 

So please feel free to address the comments in a follow-up.

ldk-server-client/src/client.rs
benthecarman: It was easier to do this way for now and since we're planning on switching it out, didn't want to complicate it

PR #182, opened 2026-04-07
Reject invalid parameters instead of silently dropping them

benthecarman:
Previously, `onchain_send` silently fell back to the node's default fee estimation when given an invalid `fee_rate_sat_per_vb` (e.g., overflows), and `bolt12_receive` silently dropped the `quantity` parameter when `amount_msat` was not set. Both cases now return an error instead.

Matching CLI-side validation added for both cases.

### DIFF
--- a/ldk-server-cli/src/main.rs
+++ b/ldk-server-cli/src/main.rs
@@ -779,6 +779,11 @@ async fn main() {
 			);
 		},
 		Commands::Bolt12Receive { description, amount, expiry_secs, quantity } => {
+			if quantity.is_some() && amount.is_none() {
+				handle_error_msg(
+					"quantity can only be set for fixed-amount offers (amount must be provided)",
+				);
+			}
 			let amount_msat = amount.map(|a| a.to_msat());
 			handle_response_result::<_, Bolt12ReceiveResponse>(
 				client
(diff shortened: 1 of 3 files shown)


ldk-server-cli/src/main.rs
```diff
@@ -634,6 +634,12 @@ async fn main() {
 			);
 		},
 		Commands::OnchainSend { address, amount, send_all, fee_rate_sat_per_vb } => {
+			if let Some(rate) = fee_rate_sat_per_vb {
+				// Mirrors FeeRate::from_sat_per_vb: sat/vb * 1000/4 = sat/kwu
```
tnull: If we get into that territory, let's just take a dependency on `bitcoin` and use `bitcoin::FeeRate`?

ldk-server-cli/src/main.rs
```diff
@@ -780,6 +780,11 @@ async fn main() {
 			);
 		},
 		Commands::Bolt12Receive { description, amount, expiry_secs, quantity } => {
+			if quantity.is_some() && amount.is_none() {
+				handle_error_msg(
+					"quantity can only be set for fixed-amount offers (amount must be provided)",
```
tnull, line 784: Debatable, @TheBlueMatt previously wanted to drop quantity from the LDK Node API (https://github.com/lightningdevkit/ldk-node/pull/642, but was closed as abandoned). Matt, do you still think we should?

ldk-server-cli/src/main.rs
benthecarman: Just removed the cli side validation. Too hit this would have to enter something greater than `u64::MAX / 250` which is super unlikely a user enters anyways and don't want to add the whole rust-bitcoin dep tree to the cli just for that.

PR #183, opened 2026-04-09
Bump LDK Node dependency to `fe692f3e397b311489aff8b2dc00761f7d10a69a`

tnull:


benthecarman:
Tests don't compile

ldk-server/ldk-server-config.toml
```diff
@@ -84,6 +84,11 @@ max_payment_size_msat = 25000000000       # 0.25 BTC
 # funding transaction until the client claimed sufficient HTLC parts to pay for the channel open.
 client_trusts_lsp = false
 
+# Allow the LSP clients to keep zero channel reserves, i.e., spend their full balance.
+# Note this will potentially allow them to attempt cheating without risking financial penalty, so
```
tankyleo, line 83: nit: I find the word "potentially" to actually decrease the communicated level of risk in this sentence. In my mind there is no "potentially", zero-reserve allow free attempts at cheating, full stop :)

Ie say you exhaust your channel balance. Then you broadcast an old state with your balance at eg. 10_000sats. The counterparty claims back the 10_000sats. You've lost no funds in this attempt (it's free). With reserves on, you would have lost 1000 sats.

tnull: > Ie say you exhaust your channel balance

Heh, right there you have you're potentiality. I know, splitting hairs, but you started :)

> Then you broadcast an old state with your balance at eg. 10_000sats. The counterparty claims back the 10_000sats. You've lost no funds in this attempt (it's free). With reserves on, you would have lost 1000 sats.

Sure, I'll drop the word if you find it makes a huge difference.

tnull:
> Tests don't compile

Should be fixed.

ldk-server/ldk-server-config.toml
benthecarman: any reason we don't just copy-paste the ldk-node docs/?

tnull: > any reason we don't just copy-paste the ldk-node docs/?

Well, for one you had outstanding nits that are might be pending to be addressed in a follow-up, as far as I understood? But, would you prefer that I just 1:1 copy the LDK Node docs?

benthecarman:
Need to add `allow_client_0reserve` to the config in the e2e tests as well

tnull:
> Need to add `allow_client_0reserve` to the config in the e2e tests as well

Ah, right.

benthecarman (approved):

PR #184, opened 2026-04-09
gRPC follow ups

benthecarman:
- Use gRPC status code consts for client: https://github.com/lightningdevkit/ldk-server/pull/176#discussion_r3057905222
 - Bias select toward shutdown in event stream: https://github.com/lightningdevkit/ldk-server/pull/176#discussion_r3057918800


tnull (approved):
Thanks!

PR #186, opened 2026-04-14
Add initial guides & documentation

benthecarman:
Add docs/ directory with guides for getting started, configuration, the gRPC API, Tor setup, and production operations. Add per-crate READMEs for ldk-server-client, ldk-server-grpc, and ldk-server-cli.

Include READMEs in lib.rs via doc = include_str! so rustdoc validates the code examples. Update the root README to link to the new docs and fix outdated REST/port references.

Some of the things documented are still subject to change (auth) but if we're able to land this first, we can then just update the docs with the changes as well.

### DIFF
--- a/docs/api-guide.md
+++ b/docs/api-guide.md
@@ -1,0 +1,240 @@
+# API Guide
+
+LDK Server exposes a gRPC API over HTTP/2 with TLS. This guide covers transport, authentication,
+and provides an index of all available RPCs. For field-level details on each request and response,
+refer to the proto definitions, which are the canonical reference and include links to the
+underlying LDK Node documentation.
+
+## Transport
+
+- **Protocol:** gRPC over HTTP/2 with TLS (self-signed by default)
+- **Default address:** `127.0.0.1:3536`
+- **Content-Type:** `application/grpc+proto`
+- **Service name:** `api.LightningNode`
+- **Full RPC path format:** `/api.LightningNode/<MethodName>`
+
+## Authentication
+
+Every gRPC request must include an `x-auth` metadata header with an HMAC-SHA256 signature:
+
+```
+x-auth: HMAC <unix_timestamp>:<hmac_hex>
+```
+
+Where:
+
+- `unix_timestamp` is the current time in seconds since the Unix epoch
+- `hmac_hex` is the hex-encoded result of `HMAC-SHA256(api_key_bytes, timestamp_be_bytes)`
+    - `api_key_bytes` is the API key string encoded as UTF-8 bytes
+    - `timestamp_be_bytes` is the timestamp as a big-endian 8-byte unsigned integer
+
+The server rejects requests where the timestamp differs from the server's clock by more than
+**60 seconds**.
+
+## TLS
+
+The server auto-generates a self-signed ECDSA P-256 certificate on first startup, stored at
+`<storage_dir>/tls.crt`. Clients must pin this certificate (not rely on system trust roots)
+since it is self-signed.
+
+For the Rust client library, pass the PEM contents to `LdkServerClient::new()`. For other
+languages, configure your gRPC channel to trust the server's certificate file.
+
+## Proto Definitions
+
+The canonical API definitions live in `ldk-server-grpc/src/proto/`:
+
+| File           | Contents                                            |
+|----------------|-----------------------------------------------------|
+| `api.proto`    | All RPC request/response messages and documentation |
+| `types.proto`  | Shared types (Payment, Channel, Peer, etc.)         |
+| `events.proto` | Event envelope and event types for streaming        |
+| `error.proto`  | Error response definitions                          |
+
+### Generating Client Stubs
+
+Any standard `protoc` toolchain can generate clients from these proto files. The proto directory
+path is `ldk-server-grpc/src/proto/`. For Rust specifically, the `ldk-server-client` crate
+provides a ready-made async client.
+
+## Error Model
+
+Errors are returned as standard gRPC status codes:
+
+| gRPC Code                 | Meaning                                                          |
+|---------------------------|------------------------------------------------------------------|
+| `INVALID_ARGUMENT` (3)    | Malformed request or invalid parameters                          |
+| `FAILED_PRECONDITION` (9) | Lightning operation error (e.g., insufficient balance, no route) |
+| `INTERNAL` (13)           | Server-side bug                                                  |
+| `UNAUTHENTICATED` (16)    | Missing or invalid `x-auth` header                               |
+
+The `grpc-message` trailer contains a human-readable error description.
+
+## Endpoint Reference
+
+All RPCs are unary (single request, single response) unless noted otherwise.
+
+### Node Information
+
+| RPC           | Description                                                                         |
+|---------------|-------------------------------------------------------------------------------------|
+| `GetNodeInfo` | Node ID, best block, sync timestamps, listening/announcement addresses, alias, URIs |
+| `GetBalances` | On-chain, Lightning channel, and claimable balance breakdown                        |
+
+### On-Chain
+
+| RPC              | Description                                                          |
+|------------------|----------------------------------------------------------------------|
+| `OnchainReceive` | Generate a new on-chain funding address                              |
+| `OnchainSend`    | Send to a Bitcoin address (with optional fee rate and send-all mode) |
+
+### BOLT11 Payments
+
+| RPC             | Description                                                       |
+|-----------------|-------------------------------------------------------------------|
+| `Bolt11Receive` | Create an invoice (fixed or variable amount) with automatic claim |
+| `Bolt11Send`    | Pay a BOLT11 invoice (with optional routing config)               |
+
+### BOLT11 Hodl Invoices
+
+These RPCs support a manual claim/fail workflow for held payments. See
+[Hodl Invoice Lifecycle](#hodl-invoice-lifecycle) below.
+
+| RPC                    | Description                                                        |
+|------------------------|--------------------------------------------------------------------|
+| `Bolt11ReceiveForHash` | Create an invoice for a given payment hash (manual claim required) |
+| `Bolt11ClaimForHash`   | Claim a held payment by providing the preimage                     |
+| `Bolt11FailForHash`    | Reject a held payment                                              |
+
+### BOLT11 JIT Channels (LSPS2)
+
+Requires an `[liquidity.lsps2_client]` configuration. The LSP opens a channel just-in-time
+when the invoice is paid.
+
+| RPC                                        | Description                                               |
+|--------------------------------------------|-----------------------------------------------------------|
+| `Bolt11ReceiveViaJitChannel`               | Create a fixed-amount invoice with JIT channel opening    |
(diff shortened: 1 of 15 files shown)


docs/api-guide.md
```diff
@@ -0,0 +1,240 @@
...
+
+LDK Server exposes a gRPC API over HTTP/2 with TLS. This guide covers transport, authentication,
+and provides an index of all available RPCs. For field-level details on each request and response,
+refer to the proto definitions, which are the canonical reference and include links to the
+underlying LDK Node documentation.
+
+## Transport
+
+- **Protocol:** gRPC over HTTP/2 with TLS (self-signed by default)
+- **Default address:** `127.0.0.1:3536`
```
tnull, line 11: In general, but especially if we want to document concrete values, do we have a plan for how to keep the API docs here in-sync and up-to-date with the actual implementation? Could we somehow check that in CI? Or would it make sense to have this live in or generated by `rustdoc`, so that we can link to the actual values in the code?

docs/configuration.md
```diff
@@ -0,0 +1,124 @@
+# Configuration
+
+LDK Server can be configured via a TOML file, environment variables, or CLI arguments.
+The [annotated config template](../ldk-server/ldk-server-config.toml) shows every available
```
tnull: Somewhat orthogonal, but should we move the TOML to a contrib folder, where we could also add a `systemd` service file going forward, for instance?

docs/configuration.md
```diff
@@ -0,0 +1,124 @@
...
+### `[liquidity.lsps2_client]`
+
+Connects to an [LSPS2](https://github.com/BitcoinAndLightningLayerSpecs/lsp/blob/main/LSPS2/README.md)
+Liquidity Service Provider for just-in-time (JIT) inbound channel opening. When configured,
+the `Bolt11ReceiveViaJitChannel` and `Bolt11ReceiveVariableAmountViaJitChannel` RPCs become
+available, the LSP will open a channel on the fly when the generated invoice is paid.
+
+Requires the LSP's public key and address. Some LSPs also require an authentication token.
+
+### `[liquidity.lsps2_service]`
```
tnull, line 100: Might be worth mentioning this also requires compilation with the experimental feature enabled.

docs/configuration.md
```diff
@@ -0,0 +1,124 @@
...
+behalf of clients. This involves setting fee parameters (opening fee, minimum fee, overprovisioning
+ratio), channel lifetime guarantees, payment size limits, and the trust model.
+
+The `client_trusts_lsp` flag controls when the funding transaction is broadcast: when enabled,
+the LSP delays broadcasting until the client has claimed enough HTLC parts to cover the
+channel opening cost.
+
+### `[metrics]`
+
+Enables a Prometheus metrics endpoint at `GET /metrics` on the gRPC port, with optional
```
tnull: nit: Link Prometheus, here and elsewhere.

docs/getting-started.md
```diff
@@ -0,0 +1,152 @@
...
+
+- **Rust** 1.85.0 or later
+- **A Bitcoin chain backend** (one of):
+    - [Bitcoin Core](https://bitcoincore.org/) (bitcoind) with RPC enabled
+    - An [Electrum](https://electrum.org/) server
+    - An [Esplora](https://github.com/Blockstream/esplora) API endpoint
+
+No other external dependencies are required.
+
+## Build
```
tnull, line 15: Might want to describe features available?

docs/operations.md
```diff
@@ -0,0 +1,181 @@
...
+
+## Process Management
+
+### systemd
+
+LDK Server integrates with systemd via `sd_notify`. It sends `READY=1` after the gRPC listener
+is bound and `STOPPING=1` when shutting down. A sample unit file:
+
+```ini
+[Unit]
```
tnull: nit: Let's just provide this in a `contrib` folder?

docs/operations.md
```diff
@@ -0,0 +1,181 @@
...
+
+The server handles `SIGTERM` and `CTRL-C` (SIGINT). On receipt, it:
+
+1. Signals all active streaming clients (SubscribeEvents) to disconnect
+2. Stops the LDK Node (persists channel state)
+3. Exits cleanly
+
+### Log Rotation
+
+The server reopens its log file on `SIGHUP`. This integrates with standard `logrotate`:
```
tnull: Hmm, this seems to be missing instructions what to do with that code/config blob below.

Should we also include a big warning note pointing out that you really need to do this, as otherwise your harddrive might be filled up over time?

docs/operations.md
```diff
@@ -0,0 +1,181 @@
...
+
+## Backups
+
+### What to Back Up
+
+| File                                   | Priority     | Description                                                                |
+|----------------------------------------|--------------|----------------------------------------------------------------------------|
+| `<storage_dir>/keys_seed`              | **Critical** | Node identity and master secret. Required to recover on-chain funds.       |
+| `<network_dir>/ldk_node_data.sqlite`   | **Critical** | Channel state and on-chain wallet data. Required to recover channel funds. |
+| `<network_dir>/ldk_server_data.sqlite` | Nice-to-have | Payment and forwarding history                                             |
```
tnull, line 53: Ugh, should we even document this to begin with?

docs/tor.md
```diff
@@ -0,0 +1,98 @@
...
+This requires a running Tor daemon with a SOCKS port. Only connections to `.onion` peers use
+the proxy. Connections to IPv4/IPv6 peers, Electrum servers, and Esplora endpoints are **not**
+routed through Tor.
+
+## Inbound Connections
+
+To make your node reachable as a Tor hidden service, you need to configure Tor itself. LDK
+Server does not manage this automatically.
+
+### 1. Configure the Hidden Service
```
tnull, line 39: Hmmm, did you test this, does the hidden service setup really work right now?

docs/tor.md
```diff
@@ -0,0 +1,98 @@
...
+
+The `[tor]` section in the config sets a SOCKS proxy for outbound connections to OnionV3 peers:
+
+```toml
+[tor]
+proxy_address = "127.0.0.1:9050"
+```
+
+This requires a running Tor daemon with a SOCKS port. Only connections to `.onion` peers use
+the proxy. Connections to IPv4/IPv6 peers, Electrum servers, and Esplora endpoints are **not**
```
tnull, line 31: It seems the Tor bypass of Electrum servers etc.  derserves a big `Warning` sign. I'm honestly even hesitant to advertise Tor support in such a general way until we're sure we're not leaking any connections.

tnull (commented):
Thanks!

docs/api-guide.md
```diff
@@ -0,0 +1,240 @@
...
+| `GraphListChannels` | List all known short channel IDs in the network graph |
+| `GraphGetChannel`   | Get channel details by short channel ID               |
+| `GraphListNodes`    | List all known node IDs in the network graph          |
+| `GraphGetNode`      | Get node details by node ID                           |
+
+### Routing
+
+| RPC                       | Description                                          |
+|---------------------------|------------------------------------------------------|
+| `ExportPathfindingScores` | Export the router's pathfinding score cache          |
```
Anyitechs, line 183: Do we also want to add that we support "importing" scorer files via `set_pathfinding_scores_source`?

docs/operations.md
```diff
@@ -0,0 +1,181 @@
...
+      - targets: [ 'localhost:3536' ]
+```
+
+### Available Metrics
+
+Metrics cover:
+
+- On-chain and Lightning balances
+- Channel count and balances
+- Payment counts (successful, failed)
```
Anyitechs: We also cover pending payments count, and channels count in different state (public, private)

Anyitechs (commented):
Looking good, thank you!

docs/tor.md
benthecarman: I tested mine, can connect to me if you want :) `0329febe813633207345f016d06d1e2b07f7b977ae7b34cda67d026769d7beb0b2@abz6grablzwec2afhng5vxtrbeioc2bczlbeeqnf57anl2lcjean3aad.onion:9735`

docs/api-guide.md
benthecarman: We could put these as docs in the rust files so we could use the `const`s but feels out of place. Otherwise, not really sure how we'd enforce because its pretty unstructured and hard to detect, maybe having claude verify?

I agree this can be a problem but most of these things should be pretty unchanging (don't see why we'd change the default port). Maybe just remove the `Endpoint Reference` section as that will have the most changes and just keep sections like `Hodl Invoice Lifecycle` and `Pagination` that just explain higher level flows rather than individual endpoints.

docs/configuration.md
benthecarman: I was planning/hoping to remove that flag so we can ship a single binary, dockerfile, etc but yeah will add that for now since we haven't done it yet.

docs/operations.md
benthecarman: This should all be in the ldk-node db eventually so we should be able to remove it. In the meantime I'd rather document it so people know what the 2 dbs are. I wouldn't want someone to mess with `ldk_node_data` thinking they don't need it because they're running `ldk-server`, not `ldk-node`

docs/tor.md
benthecarman: I'd like to include a doc for it because I know people will ask for it. I agree on the warning sign, will add. However, ideally people aren't using electrum with ldk-server and are using a proper bitcoind node so there wouldn't be leaks anyways.

docs/operations.md
benthecarman: added

docs/api-guide.md
benthecarman: We mention it briefly in the configuration.md. I didn't want to highlight individual config options too much because that can quickly become out of date and we just give descriptions about the overall sections

docs/configuration.md
tnull: Yeah, IMO we should only drop that flag once we refactored the interface to something that is more usable in actual production, not just as a PoC.

docs/api-guide.md
```diff
@@ -0,0 +1,240 @@
...
+| gRPC Code                 | Meaning                                                          |
+|---------------------------|------------------------------------------------------------------|
+| `INVALID_ARGUMENT` (3)    | Malformed request or invalid parameters                          |
+| `FAILED_PRECONDITION` (9) | Lightning operation error (e.g., insufficient balance, no route) |
+| `INTERNAL` (13)           | Server-side bug                                                  |
+| `UNAUTHENTICATED` (16)    | Missing or invalid `x-auth` header                               |
+
+The `grpc-message` trailer contains a human-readable error description.
+
+## Endpoint Reference
```
tnull, line 73: Given you're comment over on #188, how is this different/less annoying to maintain over time?

tnull (approved):
Feel free to land, can still iterate furhter.

tnull:
Actually just landing this, can continue the discussion still.

docs/api-guide.md
benthecarman: Yeah might be better to remove both

PR #187, opened 2026-04-15
Add `OpenChannel::disable_counterparty_reserve`

tankyleo:
When set, the counterparty will be allowed to spend its entire balance in the channel.

This setting cannot be set together with `announce_channel`.

### DIFF
--- a/ldk-server-cli/src/main.rs
+++ b/ldk-server-cli/src/main.rs
@@ -389,6 +389,11 @@ enum Commands {
 		push_to_counterparty: Option<Amount>,
 		#[arg(long, help = "Whether the channel should be public")]
 		announce_channel: bool,
+		#[arg(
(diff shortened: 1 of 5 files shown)


ldk-server-cli/src/main.rs
```diff
@@ -389,6 +389,11 @@ enum Commands {
 		push_to_counterparty: Option<Amount>,
 		#[arg(long, help = "Whether the channel should be public")]
 		announce_channel: bool,
+		#[arg(
+			long,
+			help = "Allow the counterparty to spend all its channel balance. This cannot be set together with `announce_channel`."
```
benthecarman, line 394: can we enforce the cannot set together with announce_channel on the cli side too

benthecarman (approved):

PR #188, opened 2026-04-15
Move `ldk-server-mcp` into the workspace

tnull:
Moving from https://github.com/tnull/ldk-server-mcp

Draft for now.


.github/workflows/mcp.yml
```diff
@@ -0,0 +1,66 @@
...
+        uses: actions/checkout@v6
+
+      - name: Install Rust stable toolchain
+        run: |
+          curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --profile=minimal --default-toolchain stable
+          rustup override set stable
+          rustup component add rustfmt clippy
+
+      - name: Check formatting
+        run: cargo fmt --all -- --check
```
benthecarman: we already check this elsewhere

.github/workflows/mcp.yml
```diff
@@ -0,0 +1,66 @@
...
+        run: |
+          source ./scripts/download_bitcoind.sh
+          mkdir -p bin
+          mv "$BITCOIND_EXE" bin/bitcoind-${{ runner.os }}-${{ runner.arch }}
+
+      - name: Set bitcoind environment variable
+        run: echo "BITCOIND_EXE=$( pwd )/bin/bitcoind-${{ runner.os }}-${{ runner.arch }}" >> "$GITHUB_ENV"
+
+      - name: Run MCP end-to-end tests
+        run: cargo test --manifest-path e2e-tests/Cargo.toml mcp --verbose --color=always -- --test-threads=4 --nocapture
```
benthecarman: pretty sure our current e2e tests job will run this already

e2e-tests/build.rs
```diff
@@ -47,4 +55,7 @@ fn main() {
 	println!("cargo:rerun-if-changed=../ldk-server-cli/Cargo.toml");
 	println!("cargo:rerun-if-changed=../ldk-server-grpc/src");
 	println!("cargo:rerun-if-changed=../ldk-server-grpc/Cargo.toml");
+	println!("cargo:rerun-if-changed=../ldk-server-mcp/src");
+	println!("cargo:rerun-if-changed=../ldk-server-mcp/tests");
```
benthecarman: the tests changing wouldn't effect the mcp binary, we don't need this here i dont think

ldk-server-mcp/src/tools/handlers.rs
```diff
@@ -0,0 +1,923 @@
...
+use ldk_server_client::ldk_server_grpc::types::{
+	bolt11_invoice_description, channel_config, Bolt11InvoiceDescription, ChannelConfig, PageToken,
+	RouteParametersConfig,
+};
+use serde_json::{json, Value};
+
+const DEFAULT_MAX_TOTAL_CLTV_EXPIRY_DELTA: u32 = 1008;
+const DEFAULT_MAX_PATH_COUNT: u32 = 10;
+const DEFAULT_MAX_CHANNEL_SATURATION_POWER_OF_HALF: u32 = 2;
+const DEFAULT_EXPIRY_SECS: u32 = 86_400;
```
benthecarman: We have these in the cli as well. Maybe worth moving into `ldk-server-grpc` or `ldk-server-client`

ldk-server-mcp/README.md
```diff
@@ -0,0 +1,179 @@
...
+        "LDK_BASE_URL": "localhost:3000",
+        "LDK_API_KEY": "your_hex_encoded_api_key",
+        "LDK_TLS_CERT_PATH": "/path/to/tls.crt"
+      }
+    }
+  }
+}
+```
+
+## Available Tools
```
benthecarman, line 90: Listing every rpc will be annoying to keep up to date

benthecarman (commented):
Did a quick skim and ran through codex

Codex Review:

```
  - [P2] Honor storage.disk.dir_path when resolving credentials — /home/ben/projects/ldk-server/ldk-server-mcp/src/
    config.rs:57-60
    If the shared ldk-server config uses a custom storage.disk.dir_path, this parser drops that section entirely, so
    resolve_config() can only look in ~/.ldk-server/... for api_key and tls.crt. In deployments that store node state
    outside the default directory (a setup ldk-server-cli already supports), the MCP bridge will fail to authenticate
    unless users also set both LDK_API_KEY and LDK_TLS_CERT_PATH by hand.
  - [P2] Fall back to the default gRPC address when no base URL is configured — /home/ben/projects/ldk-server/ldk-
    server-mcp/src/config.rs:112-116
    In setups that rely on the standard 127.0.0.1:3536 gRPC address and only use the default credential files on disk,
    this path exits with “Base URL not provided” instead of using the same default as ldk-server-cli. As a result, ldk-
    server-mcp cannot start in the zero-config case unless users create a config file or export LDK_BASE_URL, even
    though the server is running at the normal address.
```

I think these are two things we fixed in the cli since you created the MCP server.

ldk-server-mcp/README.md
tnull: > Listing every rpc will be annoying to keep up to date

Yeah, though it should be easy for any agent, and without it will also generally be annoying to add the tools and keep them in-sync with the RPC API. 

Now dropped the list.

ldk-server-mcp/src/config.rs
```diff
@@ -0,0 +1,300 @@
+// This file is Copyright its original authors, visible in version control
```
benthecarman, line 1: most of this file seems shared with logic we have in the cli. maybe also worth putting in the client.

ldk-server-mcp/src/tools/handlers.rs
```diff
@@ -0,0 +1,922 @@
...
+	bolt11_invoice_description, channel_config, Bolt11InvoiceDescription, ChannelConfig, PageToken,
+	RouteParametersConfig,
+};
+use ldk_server_client::{
+	DEFAULT_EXPIRY_SECS, DEFAULT_MAX_CHANNEL_SATURATION_POWER_OF_HALF, DEFAULT_MAX_PATH_COUNT,
+	DEFAULT_MAX_TOTAL_CLTV_EXPIRY_DELTA,
+};
+use serde_json::{json, Value};
+
+fn hex_encode(bytes: &[u8]) -> String {
```
benthecarman: we should have `hex-conservative` available here

ldk-server-mcp/src/tools/mod.rs
```diff
@@ -0,0 +1,311 @@
...
+	ToolSpec { name, description, input_schema, handler }
+}
+
+impl ToolRegistry {
+	pub fn list_tools(&self) -> Vec<&ToolDefinition> {
+		self.tools.iter().map(|(def, _)| def).collect()
+	}
+
+	pub async fn call_tool(
+		&self, client: &LdkServerClient, name: &str, args: Value,
```
benthecarman, line 54: Using `serde_json::Value` everywhere makes things feel really flaky. I think it'd make sense to bring back the json deserializers for the request types and then use those. That would make things a lot more type safe and less error prone.

ldk-server-mcp/src/tools/handlers.rs
```diff
@@ -0,0 +1,922 @@
...
+	}
+	let index = parts[1].parse::<i64>().map_err(|_| "Invalid page token index".to_string())?;
+	Ok(PageToken { token: parts[0].to_string(), index })
+}
+
+fn format_page_token(pt: &PageToken) -> String {
+	format!("{}:{}", pt.token, pt.index)
+}
+
+fn build_route_parameters(args: &Value) -> RouteParametersConfig {
```
benthecarman: this file would be hugely cleaned up if we used the proto struct's json deserialzers

ldk-server-mcp/src/main.rs
```diff
@@ -0,0 +1,140 @@
...
+
+	let cfg = match config::resolve_config(config_path) {
+		Ok(cfg) => cfg,
+		Err(e) => {
+			eprintln!("Error: {e}");
+			std::process::exit(1);
+		},
+	};
+
+	let client = match LdkServerClient::new(cfg.base_url, cfg.api_key, &cfg.tls_cert_pem) {
```
benthecarman, line 55: should we call `get_info` to start just to make sure we're connected?

ldk-server-mcp/src/main.rs
```diff
@@ -0,0 +1,140 @@
...
+
+		// Notifications have no id — do not respond
+		if request.id.is_none() {
+			continue;
+		}
+
+		let id = request.id.unwrap();
+
+		let response_str = match request.method.as_str() {
+			"initialize" => {
```
benthecarman, line 117: can we make these `const` s

ldk-server-mcp/src/tools/mod.rs
```diff
@@ -0,0 +1,311 @@
...
+	) -> ToolCallResult {
+		for (def, handler) in &self.tools {
```
benthecarman: Would be better to have this as a `HashMap` keyed by the name so we don't have to iterate through every time.

ldk-server-mcp/src/tools/mod.rs
```diff
@@ -0,0 +1,311 @@
...
+			if def.name == name {
+				return match handler(client, args).await {
+					Ok(value) => {
+						let text = serde_json::to_string_pretty(&value)
```
benthecarman: Why pretty print it here? This is just going to the agent, so wouldnt it waste tokens

ldk-server-mcp/src/mcp.rs
```diff
@@ -0,0 +1,89 @@
...
+// This file is licensed under the Apache License, Version 2.0 <LICENSE-APACHE
+// or http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your option.
+// You may not use this file except in accordance with one or both of these
+// licenses.
+
+use serde::Serialize;
+use serde_json::Value;
+
+pub const PROTOCOL_VERSION: &str = "2024-11-05";
```
benthecarman: The latest version is `2025-11-25`, might be worth updating too. With my brief chat with claude about the differences that'd effect us is that the error handling is slightly improved and more metadata. It also has the remote mcp stuff using SSE but we can ignore that for now.

ldk-server-mcp/src/main.rs
```diff
@@ -0,0 +1,140 @@
...
+			},
+			"tools/list" => {
+				let tools = registry.list_tools();
+				let resp = JsonRpcResponse::new(id, serde_json::json!({ "tools": tools }));
+				serde_json::to_string(&resp).unwrap()
+			},
+			"tools/call" => {
+				let params = request.params.unwrap_or(Value::Null);
+				let tool_name = params.get("name").and_then(|v| v.as_str()).unwrap_or("");
+				let tool_args = params.get("arguments").cloned().unwrap_or(serde_json::json!({}));
```
benthecarman: instead of the `upwrap_or`s here we should be returning a `INVALID_PARAMS` error

ldk-server-mcp/src/tools/handlers.rs
```diff
@@ -0,0 +1,922 @@
...
+		.unwrap_or(DEFAULT_EXPIRY_SECS);
+
+	let response = client
+		.bolt11_receive(Bolt11ReceiveRequest {
+			description: invoice_description,
+			expiry_secs,
+			amount_msat,
+		})
+		.await
+		.map_err(|e| e.message.clone())?;
```
benthecarman: Here and a lot of others, you don't need to `clone` the error message

benthecarman: Instead it actually would make sense to make a converter for `LdkServerError` to a dedicated error type here. Then we could handle things better like if they server marked the params as invalid vs an actual server error. Then we could actually use `INTERNAL_ERROR` too

benthecarman (commented):
needs rebase too

ldk-server-mcp/src/tools/handlers.rs
```diff
@@ -0,0 +1,348 @@
...
+	let request: DecodeOfferRequest = parse_request(args)?;
+	let response = client.decode_offer(request).await.map_err(McpError::from)?;
+	serialize_response(response)
+}
+
+// `message` is a proto `bytes` field, which deserializes as a numeric array rather than a UTF-8
+// string, so we manually extract the string and turn it into bytes instead of delegating to the
+// proto-typed deserializer.
+pub async fn handle_sign_message(client: &LdkServerClient, args: Value) -> Result<Value, McpError> {
+	let message = args
```
benthecarman: any reason we aren't using the json decoder here? and in `handle_verify_signature`

ldk-server-mcp/CLAUDE.md
```diff
@@ -0,0 +1,71 @@
...
+    handlers.rs  — Handler functions: JSON args -> ldk-server-client call -> JSON result
+```
+
+## MCP Protocol
+
+- **Version**: `2025-11-25`
+- **Spec**: https://spec.modelcontextprotocol.io/
+- **Transport**: stdio (one JSON-RPC 2.0 message per line)
+- **Methods implemented**: `initialize`, `tools/list`, `tools/call`
+- **Notifications handled**: `notifications/initialized` (ignored, no response)
```
benthecarman, line 39: I don't see us handling this anywhere, looks like we'd return `METHOD_NOT_FOUND`

tnull: No, we just skip in `main`:

```rust

		// Notifications have no id — do not respond
		if request.id.is_none() {
			continue;
		}
```

ldk-server-mcp/src/tools/handlers.rs
tnull: Hmm, see the comment above. Now introduced some extra `{SignMessage,VerifySignature}Args` structs deriving `Deserialize` to make this a bit cleaner. Let me know what you think.

benthecarman:
Can you squash?

tnull:
> Can you squash?

ldk-server-mcp/src/tools/handlers.rs
```diff
@@ -0,0 +1,342 @@
...
+	let response = client.bolt11_receive_via_jit_channel(request).await.map_err(McpError::from)?;
+	serialize_response(response)
+}
+
+pub async fn handle_bolt11_receive_variable_amount_via_jit_channel(
+	client: &LdkServerClient, args: Value,
+) -> Result<Value, McpError> {
+	let mut request: Bolt11ReceiveVariableAmountViaJitChannelRequest = parse_request(args)?;
+	if request.expiry_secs == 0 {
+		request.expiry_secs = DEFAULT_EXPIRY_SECS;
```
benthecarman, line 175: I think we'll need to do similar logic for `DEFAULT_MAX_TOTAL_CLTV_EXPIRY_DELTA`, `DEFAULT_MAX_PATH_COUNT`, and `DEFAULT_MAX_CHANNEL_SATURATION_POWER_OF_HALF` where relevant

benthecarman (commented):
final thing and then i think this is good

benthecarman (commented):
lgtm on squash!

Image

tnull:
> lgtm on squash!

benthecarman:
the proto ci is failing now?

tnull:
> the proto ci is failing now?

Ah, silent merge conflict it seems. Rebased.

benthecarman (approved):
amazing!

PR #189, opened 2026-04-15
Implement configurable native size/time log rotation & deletion

Anyitechs:
This introduces a configurable native size/time log rotation & deletion functionality. It updates the `ServerLogger` to track file size and age internally, triggering rotation at a default size of 50MB or 24 hours, if the config values `max_size_mb`/`rotation_interval_hours` are unset. We keep only `max_files` (defaults to 5) rotated log files at a time, deleting older log files.

We also now allow users to turn off logging to file, but the default is logging to file and `stdout`/`stderr`. Internal rotation can be disabled for file logging by setting the `max_size_mb` and `rotation_interval_hours` params to `0`

### DIFF
--- a/contrib/ldk-server-config.toml
+++ b/contrib/ldk-server-config.toml
@@ -15,6 +15,10 @@ dir_path = "/tmp/ldk-server/"                 # Path for LDK and BDK data persis
 [log]
 level = "Debug"                               # Log level (Error, Warn, Info, Debug, Trace)
 #file = "/tmp/ldk-server/ldk-server.log"      # Log file path
+log_to_file = true                           # Enable logging to a file (default: true, also logs to both stdout and stderr)
+#max_size_mb = 50                              # Max size of log file before rotation (default: 50MB)
+#rotation_interval_hours = 24                  # Max age of log file before rotation (default: 24h)
+#max_files = 5                                 # Number of rotated log files to keep (default: 5)
 
 [tls]
 #cert_path = "/path/to/tls.crt"               # Path to TLS certificate, by default uses dir_path/tls.crt
--- a/contrib/ldk-server-config.toml
+++ b/contrib/ldk-server-config.toml
@@ -15,6 +15,14 @@ dir_path = "/tmp/ldk-server/"                 # Path for LDK and BDK data persis
 [log]
 level = "Debug"                               # Log level (Error, Warn, Info, Debug, Trace)
 #file = "/tmp/ldk-server/ldk-server.log"      # Log file path
+# Enabling `log_to_file` writes logs to the configured file while still keeping 
+# the `stdout`/`stderr` logs available too. Logs files are automatically rotated at # `max_size_mb` or `rotation_interval_hours`.
+#
+# To disable the internal rotation and keep logging to file, set `max_size_mb` # and `rotation_interval_hours` params to `0`.
+log_to_file = true                            # Enable logging to a file (default: true, also logs to both stdout and stderr)
+#max_size_mb = 50                             # Max size of log file before rotation (default: 50MB)
+#rotation_interval_hours = 24                 # Max age of log file before rotation (default: 24h)
+#max_files = 5                                # Number of rotated log files to keep (default: 5)
 
 [tls]
 #cert_path = "/path/to/tls.crt"               # Path to TLS certificate, by default uses dir_path/tls.crt
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -61,8 +61,12 @@ Where persistent data is stored. Defaults to `~/.ldk-server/` on Linux and
 
 ### `[log]`
 
-Log level and file path. The server reopens the log file on `SIGHUP`, which integrates with
-standard `logrotate` setups.
+Controls logging behavior. By default, `log_to_file` is `true` and logs are also written 
+to `stdout`/`stderr`.
+
+If `log_to_file` is enabled, the server performs internal rotation and retention 
+based on `max_size_mb`, `rotation_interval_hours`, and `max_files`. The server will 
+also reopen the log file on `SIGHUP` for compatibility with external tools like `logrotate`.
 
 ### `[tls]`
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -61,8 +61,16 @@ Where persistent data is stored. Defaults to `~/.ldk-server/` on Linux and
 
 ### `[log]`
 
-Log level and file path. The server reopens the log file on `SIGHUP`, which integrates with
-standard `logrotate` setups.
+Controls logging behavior. By default, `log_to_file` is `true` and logs are also written 
+to `stdout`/`stderr`.
+
+If `log_to_file` is enabled, logs are written to the configured file while still keeping 
+the `stdout`/`stderr` logs available too. Logs files are automatically rotated at 
+`max_size_mb` or `rotation_interval_hours`. To disable the internal rotation and keep 
+logging to file, set `max_size_mb` and `rotation_interval_hours` params to `0`.
+
+The server will also reopen the log file on `SIGHUP` for compatibility with external 
+tools like `logrotate`.
 
 ### `[tls]`
 
--- a/docs/operations.md
+++ b/docs/operations.md
@@ -21,13 +21,17 @@ The server handles `SIGTERM` and `CTRL-C` (SIGINT). On receipt, it:
 
 ### Log Rotation
 
-> **Important:** LDK Server does not rotate or truncate its own log file. Without log rotation
-> configured, the log file will grow indefinitely and can eventually fill your disk. A full
-> disk can prevent the node from persisting channel state, risking fund loss.
+By default, LDK Server logs to `stdout`/`stderr`. When running under `systemd` or Docker, 
+this allows the environment (e.g., `journald`) to handle persistence, rotation, and 
+compression automatically.
 
-The server reopens its log file on `SIGHUP`. This integrates with standard `logrotate`. Save
-the following config to `/etc/logrotate.d/ldk-server` (adjust the log path to match your
-setup):
+If you enable `log_to_file` in the configuration, LDK Server will automatically rotate 
+logs when they exceed 50MB or 24 hours (configurable) and keep the last 5 uncompressed 
+log files.
+
+If you prefer to use system `logrotate` for file logs, the server still reopens its log 
+file on `SIGHUP`. Save the following config to `/etc/logrotate.d/ldk-server` 
+(adjust the log path to match your setup):
 
 ```
 /var/lib/ldk-server/regtest/ldk-server.log {
--- a/docs/operations.md
+++ b/docs/operations.md
@@ -21,13 +21,17 @@ The server handles `SIGTERM` and `CTRL-C` (SIGINT). On receipt, it:
 
 ### Log Rotation
 
-> **Important:** LDK Server does not rotate or truncate its own log file. Without log rotation
-> configured, the log file will grow indefinitely and can eventually fill your disk. A full
-> disk can prevent the node from persisting channel state, risking fund loss.
+By default, LDK Server logs to `stdout`/`stderr` and also to file. When running under `systemd` or Docker, 
+this allows the environment (e.g., `journald`) to handle persistence, rotation, and 
+compression automatically.
 
-The server reopens its log file on `SIGHUP`. This integrates with standard `logrotate`. Save
-the following config to `/etc/logrotate.d/ldk-server` (adjust the log path to match your
-setup):
+If you enable `log_to_file` in the configuration, LDK Server writes logs to the configured file while still 
+keeping the `stdout`/`stderr` logs available too. Logs files are automatically rotated at `max_size_mb` or `rotation_interval_hours`, and the last `max_files` uncompressed log files are retained. But you can disable 
+the internal rotation and keep logging to file by setting the `max_size_mb` and `rotation_interval_hours` 
+params to `0`.
+
+If you prefer to use system `logrotate` for file logs, the server still reopens its log file on `SIGHUP`. Save 
+the following config to `/etc/logrotate.d/ldk-server` (adjust the log path to match your setup):
 
 ```
 /var/lib/ldk-server/regtest/ldk-server.log {
--- a/ldk-server/src/main.rs
+++ b/ldk-server/src/main.rs
@@ -48,7 +48,7 @@ use crate::io::persist::{
 };
 use crate::service::NodeService;
 use crate::util::config::{load_config, ArgsConfig, ChainSource};
-use crate::util::logger::ServerLogger;
+use crate::util::logger::{LogConfig, ServerLogger};
 use crate::util::metrics::Metrics;
 use crate::util::proto_adapter::{forwarded_payment_to_proto, payment_to_proto};
 use crate::util::systemd;
@@ -121,10 +121,20 @@ fn main() {
 		std::process::exit(-1);
 	}
 
-	if let Err(e) = ServerLogger::init(config_file.log_level, &log_file_path) {
-		eprintln!("Failed to initialize logger: {e}");
-		std::process::exit(-1);
-	}
+	let log_config = LogConfig {
+		log_to_file: config_file.log_to_file,
+		log_max_files: config_file.log_max_files,
+		log_max_size_bytes: config_file.log_max_size_bytes,
+		log_rotation_interval_secs: config_file.log_rotation_interval_secs,
+	};
+
+	let logger = match ServerLogger::init(config_file.log_level, &log_file_path, log_config) {
+		Ok(logger) => logger,
+		Err(e) => {
+			eprintln!("Failed to initialize logger: {e}");
+			std::process::exit(-1);
+		},
+	};
 
 	let api_key = match load_or_generate_api_key(&network_dir) {
 		Ok(key) => key,
@@ -252,6 +262,14 @@ fn main() {
 	}
 
 	runtime.block_on(async {
+		// Register SIGHUP handler for log rotation
+		let mut sighup_stream = match tokio::signal::unix::signal(SignalKind::hangup()) {
+			Ok(stream) => stream,
+			Err(e) => {
+				error!("Failed to register SIGHUP handler: {e}");
+				std::process::exit(-1);
+			}
+		};
 
 		let mut sigterm_stream = match tokio::signal::unix::signal(SignalKind::terminate()) {
 			Ok(stream) => stream,
@@ -506,6 +524,12 @@ fn main() {
 					info!("Received CTRL-C, shutting down..");
 					let _ = shutdown_tx.send(true);
 					break;
+				}
+				_ = sighup_stream.recv() => {
+					info!("Received SIGHUP, reopening log file..");
+					if let Err(e) = logger.reopen() {
+						error!("Failed to reopen log file on SIGHUP: {e}");
+					}
 				}
 				_ = sigterm_stream.recv() => {
 					info!("Received SIGTERM, shutting down..");
--- a/ldk-server/src/main.rs
+++ b/ldk-server/src/main.rs
@@ -121,13 +121,10 @@ fn main() {
 		std::process::exit(-1);
 	}
 
-	let logger = match ServerLogger::init(config_file.log_level, &log_file_path) {
-		Ok(logger) => logger,
-		Err(e) => {
-			eprintln!("Failed to initialize logger: {e}");
-			std::process::exit(-1);
-		},
-	};
+	if let Err(e) = ServerLogger::init(config_file.log_level, &log_file_path) {
+		eprintln!("Failed to initialize logger: {e}");
+		std::process::exit(-1);
+	}
 
 	let api_key = match load_or_generate_api_key(&network_dir) {
 		Ok(key) => key,
@@ -255,14 +252,6 @@ fn main() {
 	}
 
 	runtime.block_on(async {
-		// Register SIGHUP handler for log rotation
-		let mut sighup_stream = match tokio::signal::unix::signal(SignalKind::hangup()) {
-			Ok(stream) => stream,
-			Err(e) => {
-				error!("Failed to register SIGHUP handler: {e}");
-				std::process::exit(-1);
-			}
-		};
 
 		let mut sigterm_stream = match tokio::signal::unix::signal(SignalKind::terminate()) {
 			Ok(stream) => stream,
@@ -517,11 +506,6 @@ fn main() {
 					info!("Received CTRL-C, shutting down..");
 					let _ = shutdown_tx.send(true);
 					break;
-				}
-				_ = sighup_stream.recv() => {
-					if let Err(e) = logger.reopen() {
-						error!("Failed to reopen log file on SIGHUP: {e}");
-					}
 				}
 				_ = sigterm_stream.recv() => {
 					info!("Received SIGTERM, shutting down..");
--- a/ldk-server/src/main.rs
+++ b/ldk-server/src/main.rs
@@ -48,7 +48,7 @@ use crate::io::persist::{
 };
 use crate::service::NodeService;
 use crate::util::config::{load_config, ArgsConfig, ChainSource};
-use crate::util::logger::ServerLogger;
+use crate::util::logger::{LogConfig, ServerLogger};
 use crate::util::metrics::Metrics;
 use crate::util::proto_adapter::{forwarded_payment_to_proto, payment_to_proto};
 use crate::util::systemd;
@@ -110,18 +110,29 @@ fn main() {
 		Network::Regtest => storage_dir.join("regtest"),
 	};
 
-	let log_file_path = config_file.log_file_path.map(PathBuf::from).unwrap_or_else(|| {
-		let mut default_log_path = network_dir.clone();
-		default_log_path.push("ldk-server.log");
-		default_log_path
-	});
+	let log_file_path = if config_file.log_to_file {
+		let path = config_file.log_file_path.map(PathBuf::from).unwrap_or_else(|| {
+			let mut default_log_path = network_dir.clone();
+			default_log_path.push("ldk-server.log");
+			default_log_path
+		});
 
-	if log_file_path == storage_dir || log_file_path == network_dir {
-		eprintln!("Log file path cannot be the same as storage directory path.");
-		std::process::exit(-1);
-	}
+		if path == storage_dir || path == network_dir {
+			eprintln!("Log file path cannot be the same as storage directory path.");
+			std::process::exit(-1);
+		}
+		Some(path)
+	} else {
+		None
+	};
 
-	let logger = match ServerLogger::init(config_file.log_level, &log_file_path) {
+	let log_config = LogConfig {
+		log_max_files: config_file.log_max_files,
+		log_max_size_bytes: config_file.log_max_size_bytes,
+		log_rotation_interval_secs: config_file.log_rotation_interval_secs,
+	};
+
+	let logger = match ServerLogger::init(config_file.log_level, log_file_path, log_config) {
 		Ok(logger) => logger,
 		Err(e) => {
 			eprintln!("Failed to initialize logger: {e}");
@@ -519,6 +530,7 @@ fn main() {
 					break;
 				}
 				_ = sighup_stream.recv() => {
+					info!("Received SIGHUP, reopening log file..");
 					if let Err(e) = logger.reopen() {
 						error!("Failed to reopen log file on SIGHUP: {e}");
 					}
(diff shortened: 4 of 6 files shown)


benthecarman:
I would rather keep the old version. I like all the handling for the different signals in `main.rs`, makes things easier to reason about imo.

Anyitechs:
> I would rather keep the old version. I like all the handling for the different signals in `main.rs`, makes things easier to reason about imo.

As we discussed elsewhere, this is now dropped entirely as we're moving to doing the rotation ourselves rather than relying on the user to configure `logrotate`.

ldk-server/src/util/logger.rs
```diff
@@ -8,12 +8,26 @@
...
 use std::sync::{Arc, Mutex};
+use std::thread;
+use std::time::SystemTime;
 
 use log::{Level, LevelFilter, Log, Metadata, Record};
 
+/// Maximum size of the log file before it gets rotated (50 MB)
+const MAX_LOG_SIZE_BYTES: usize = 50 * 1024 * 1024;
+/// Maximum age of the log file before it gets rotated (24 hours)
+const ROTATION_INTERVAL_SECS: u64 = 24 * 60 * 60;
```
benthecarman: these should be configurable

ldk-server/src/util/logger.rs
```diff
@@ -52,30 +66,60 @@ impl ServerLogger {
...
-				*file = new_file;
-				Ok(())
+		state.file = BufWriter::new(new_file);
+
+		// Reset our rotation triggers for the new file
+		state.bytes_written = 0;
+		state.created_at = SystemTime::now();
+
+		// Spawn independent OS thread to compress the old file using native gzip
+		thread::spawn(move || match Command::new("gzip").arg("-f").arg(&rotated_path).status() {
```
benthecarman: This is assuming they have gzip installed and we have permissions to run it. I would rather just have the rust code do this

benthecarman (changes requested):

ldk-server/src/util/logger.rs
TheBlueMatt: It doesn't really seem worth taking a gzip dependency just to do log rotation. We can just rotate and delete as-is and tell people to use the system log (which handles this properly) otherwise. Really we shouldn't be doing logging ourselves anyway - log should go to stdout/err and systemd or some other logic should figure out what to do with it.

Anyitechs: I'm happy to drop the compression so we do a rotate and delete. If that works for everyone, I'll update this to introduce a configurable `max_rotated_files` that will default to the last 5 uncompressed files. Does that work?

benthecarman: I'm in favor of just doing deletion instead of compression but I would really rather keep the log file and not rely on just stdout/stderr. If we don't have an easy log file and rely on the user's systemd/docker/whatever to do and persist the logs, we will likely not have logs for lots of users and make it much harder to do support.

tnull: > I'm in favor of just doing deletion instead of compression but I would really rather keep the log file and not rely on just stdout/stderr. If we don't have an easy log file and rely on the user's systemd/docker/whatever to do and persist the logs, we will likely not have logs for lots of users and make it much harder to do support.

I'm also leaning towards just not logging to file. However I see your point that it might lead to nothing being logged at all. In that case we should however at the very least add the option to disable logging to file, and also update our docs and checked-in systemd `ldk-server.service` to use that mode, so that everybody who runs it that way will just use `journald` rather than our custom filesystem logger.

benthecarman: Yeah I like that

Anyitechs: Just added an option to disable logging to file by default and a configurable native rotate and delete here [b54e0a5](https://github.com/lightningdevkit/ldk-server/commit/b54e0a5669fd5eeb39f18e4730edcaaeadc3abb5)

contrib/ldk-server-config.toml
```diff
@@ -15,6 +15,10 @@ dir_path = "/tmp/ldk-server/"                 # Path for LDK and BDK data persis
 [log]
 level = "Debug"                               # Log level (Error, Warn, Info, Debug, Trace)
 #file = "/tmp/ldk-server/ldk-server.log"      # Log file path
+log_to_file = false                           # Enable logging to a file (default: false, logs to stderr only)
```
benthecarman: We should default to having this true. Also we log to both stdout and stderr

docs/configuration.md
```diff
@@ -61,8 +61,12 @@ Where persistent data is stored. Defaults to `~/.ldk-server/` on Linux and
...
 ### `[log]`
 
-Log level and file path. The server reopens the log file on `SIGHUP`, which integrates with
-standard `logrotate` setups.
+Controls logging behavior. By default, `log_to_file` is `false` and logs are written 
+to `stdout`/`stderr`.
+
+If `log_to_file` is enabled, the server performs internal rotation and retention 
+based on `max_size_mb`, `rotation_interval_hours`, and `max_files`. The server still 
+reopens the log file on `SIGHUP` for compatibility with external tools.
```
benthecarman:
```suggestion
If `log_to_file` is enabled, the server performs internal rotation and retention 
based on `max_size_mb`, `rotation_interval_hours`, and `max_files`. The server will 
also reopen the log file on `SIGHUP` for compatibility with external tools like `logrotate`.
```

ldk-server/src/main.rs
```diff
@@ -254,14 +258,6 @@ fn main() {
 	}
 
 	runtime.block_on(async {
-		// Register SIGHUP handler for log rotation
```
benthecarman, line 264: looks like you actually removed the `SIGHUP` handling even though in the docs you say we have it

benthecarman (changes requested):

ldk-server/src/main.rs
Anyitechs: Ah, good catch. Thanks!

Anyitechs:
Addressed review comment here [42cd66a](https://github.com/lightningdevkit/ldk-server/commit/42cd66a4d2e6838f4c2b643e81a266bbe4ccca8d) and rebased to fix conflict.

contrib/ldk-server-config.toml
```diff
@@ -15,6 +15,10 @@ dir_path = "/tmp/ldk-server/"                 # Path for LDK and BDK data persis
 [log]
 level = "Debug"                               # Log level (Error, Warn, Info, Debug, Trace)
 #file = "/tmp/ldk-server/ldk-server.log"      # Log file path
+log_to_file = true                           # Enable logging to a file (default: true, also logs to both stdout and stderr)
```
tnull: nit: Alignment of comments is off.

docs/operations.md
```diff
@@ -21,13 +21,17 @@ The server handles `SIGTERM` and `CTRL-C` (SIGINT). On receipt, it:
 
 ### Log Rotation
 
-> **Important:** LDK Server does not rotate or truncate its own log file. Without log rotation
-> configured, the log file will grow indefinitely and can eventually fill your disk. A full
-> disk can prevent the node from persisting channel state, risking fund loss.
+By default, LDK Server logs to `stdout`/`stderr`. When running under `systemd` or Docker,
```
tnull: Hmm, here we say that by default we log to stdout/stderr, but above we say `log_to_file` defaults to `true`? Probably both somewhat true, but is a bit confusing. Might be good to clarify.

ldk-server/src/util/logger.rs
```diff
@@ -158,6 +244,29 @@ fn open_log_file(log_file_path: &Path) -> Result<File, io::Error> {
 	OpenOptions::new().create(true).append(true).open(log_file_path)
 }
 
+fn cleanup_old_logs(log_file_path: &Path, max_files: usize) -> io::Result<()> {
+	let parent = log_file_path.parent().unwrap_or_else(|| Path::new("."));
```
tnull: Hmm, why are we falling back to `.`? In general, maybe it would be better to just error out if something is unexpected instead of doing the `unwrap_or` fallbacks here and below, just to be sure we never end up removing some unrelated files?

docs/configuration.md
```diff
@@ -61,8 +61,12 @@ Where persistent data is stored. Defaults to `~/.ldk-server/` on Linux and
...
 ### `[log]`
 
-Log level and file path. The server reopens the log file on `SIGHUP`, which integrates with
-standard `logrotate` setups.
+Controls logging behavior. By default, `log_to_file` is `true` and logs are also written 
+to `stdout`/`stderr`.
+
+If `log_to_file` is enabled, the server performs internal rotation and retention 
+based on `max_size_mb`, `rotation_interval_hours`, and `max_files`. The server will 
+also reopen the log file on `SIGHUP` for compatibility with external tools like `logrotate`.
```
tnull: Hmm, it seems we're now no longer compatible with `logrotate`, but rather rotate ourselves when `log_to_file` is configured? I guess that's fine, but we might want to update docs here and elsewhere?

ldk-server/src/util/logger.rs
```diff
@@ -89,52 +166,61 @@ impl Log for ServerLogger {
...
+					}
+
+					let _ = writeln!(state.file, "{}", log_line);
+					state.bytes_written += log_bytes;
+				}
 			}
 		}
 	}
 
 	fn flush(&self) {
```
tnull, line 234: Codex:
> - P1 File logs are now buffered with BufWriter (ldk-server/src/util/logger.rs:83), but normal writes do not flush (ldk-server/src/util/logger.rs:184) and shutdown does not call log::logger().flush() after the final log line (ldk-server/src/main.rs:527). Since the logger is installed
    globally, recent file logs from short runs or shutdown can remain in memory and never reach disk.

ldk-server/src/util/logger.rs
tnull: Codex has the same concern:

> - P3 Retention cleanup matches every file whose name starts with the active log basename (ldk-server/src/util/logger.rs:227). That can delete unrelated files such as ldk-server.log.backup or config artifacts once the count exceeds max_files; it should match the exact rotation filename
    pattern.

docs/configuration.md
Anyitechs: > it seems we're now no longer compatible with `logrotate`

We still are, I added that back in [42cd66a](https://github.com/lightningdevkit/ldk-server/commit/42cd66a4d2e6838f4c2b643e81a266bbe4ccca8d).

> but rather rotate ourselves when log_to_file is configured?

Yes, we rotate ourselves when `log_to_file` is configured, but also have the `SIGHUP` signal handler for users relying on `logrotate`

tnull: But how would the user disable our internal log rotation if they use `logrotate` themselves?

Anyitechs: Thanks for the catch. Addressed this in the fixup to decouple our internal log rotation from the `SIGHUP` signal handler for users relying on `logrotate`.

ldk-server/src/util/config.rs
```diff
@@ -249,6 +262,22 @@ impl ConfigBuilder {
...
+
+		if let Some(log_rotation_interval_hours) = args.log_rotation_interval_hours {
+			self.log_rotation_interval_hours = Some(log_rotation_interval_hours);
+		}
+
+		if let Some(log_max_files) = args.log_max_files {
+			self.log_max_files = Some(log_max_files);
+		}
+
+		if args.log_to_file {
```
benthecarman: since this defaults to true, there is no actual way to set `log_to_file` to false. should also make `args.log_to_file` an `Option`

ldk-server/src/main.rs
```diff
@@ -110,18 +110,30 @@ fn main() {
 		Network::Regtest => storage_dir.join("regtest"),
 	};
 
-	let log_file_path = config_file.log_file_path.map(PathBuf::from).unwrap_or_else(|| {
-		let mut default_log_path = network_dir.clone();
-		default_log_path.push("ldk-server.log");
-		default_log_path
-	});
+	let log_file_path = if config_file.log_to_file || config_file.log_file_path.is_some() {
```
benthecarman: need to remove `|| config_file.log_file_path.is_some()` from this if statement, otherwise we'll only get a log file if we configure a path, can never get to the default option

ldk-server/src/util/logger.rs
```diff
@@ -158,6 +249,45 @@ fn open_log_file(log_file_path: &Path) -> Result<File, io::Error> {
...
+			if let Some(name) = entry.file_name().to_str() {
+				if let Some(suffix) = name.strip_prefix(&rotated_prefix) {
+					// The timestamp format "YYYY-MM-DDTHH-MM-SSZ" is 20 chars, contains 'T', and ends with 'Z'
+					suffix.len() == 20 && suffix.contains('T') && suffix.ends_with('Z')
+				} else {
+					false
+				}
+			} else {
+				false
+			}
```
benthecarman: doing all these `if let Some(_)`s makes this really verbose. 
```suggestion
			entry
				.file_name()
				.to_str()
				.and_then(|name| {
					name.strip_prefix(&rotated_prefix).map(|suffix| {
						// The timestamp format "YYYY-MM-DDTHH-MM-SSZ" is 20 chars, contains 'T', and ends with 'Z'
						suffix.len() == 20 && suffix.contains('T') && suffix.ends_with('Z')
					})
				})
				.unwrap_or(false)
```

ldk-server/src/util/config.rs
```diff
@@ -358,6 +387,11 @@ impl ConfigBuilder {
 			.transpose()?
 			.unwrap_or(LevelFilter::Debug);
 
+		let log_max_size_bytes = self.log_max_size_mb.unwrap_or(50) * 1024 * 1024;
+		let log_rotation_interval_secs = self.log_rotation_interval_hours.unwrap_or(24) * 60 * 60;
+		let log_max_files = self.log_max_files.unwrap_or(5);
```
benthecarman: can we have these defaults be constants

ldk-server/src/util/logger.rs
```diff
@@ -8,74 +8,153 @@
...
 use log::{Level, LevelFilter, Log, Metadata, Record};
 
+struct LoggerState {
+	file: BufWriter<File>,
+	bytes_written: usize,
+	created_at: SystemTime,
+	log_max_size_bytes: usize,
+	log_rotation_interval_secs: u64,
+	log_max_files: usize,
+	internal_rotation_enabled: bool,
```
benthecarman: We have `internal_rotation_enabled` but I don't see anyway to actually disable the rotation. Below its set to `log_to_file` so there is probably a bug here, either way its very confusing.

benthecarman: Either way if we are going to keep the SIGHUP handling, we should have a way to disable the log rotation because otherwise the 2 systems would be fighting over each other. Made have it so if you set the param to `0` we don't rotate on that metric?

benthecarman (changes requested):

ldk-server/src/main.rs
Anyitechs: Done

ldk-server/src/util/logger.rs
Anyitechs: Done. Thank you

ldk-server/src/util/config.rs
Anyitechs: Done

ldk-server/src/util/logger.rs
Anyitechs: Ah true, thank you. I've adopted your suggestion and dropped the `internal_rotation_enabled` field entirely. We now allow disabling internal rotation by setting the `log_max_size_bytes` and `log_rotation_interval_secs` to `0` for users who want to keep logging to file but have something else do the rotation for them.

docs/operations.md
```diff
@@ -21,13 +21,17 @@ The server handles `SIGTERM` and `CTRL-C` (SIGINT). On receipt, it:
...
+By default, LDK Server logs to `stdout`/`stderr` and also to file. When running under `systemd` or Docker, 
+this allows the environment (e.g., `journald`) to handle persistence, rotation, and 
+compression automatically.
 
-The server reopens its log file on `SIGHUP`. This integrates with standard `logrotate`. Save
-the following config to `/etc/logrotate.d/ldk-server` (adjust the log path to match your
-setup):
+If you enable `log_to_file` in the configuration, LDK Server will automatically rotate 
+logs when they exceed 50MB or 24 hours (configurable) and keep the last 5 uncompressed 
+log files.
```
benthecarman: we should document that setting the 2 different configs to 0 disables that kind of rotation

contrib/ldk-server-config.toml
```diff
@@ -15,6 +15,10 @@ dir_path = "/tmp/ldk-server/"                 # Path for LDK and BDK data persis
 [log]
 level = "Debug"                               # Log level (Error, Warn, Info, Debug, Trace)
 #file = "/tmp/ldk-server/ldk-server.log"      # Log file path
+log_to_file = true                            # Enable logging to a file (default: true, also logs to both stdout and stderr)
+#max_size_mb = 50                             # Max size of log file before rotation (default: 50MB)
+#rotation_interval_hours = 24                 # Max age of log file before rotation (default: 24h)
```
benthecarman, line 24: same here

ldk-server/src/util/logger.rs
```diff
@@ -89,52 +165,65 @@ impl Log for ServerLogger {
...
+
+				if let Ok(mut state) = state_mutex.lock() {
+					let mut needs_rotation = false;
+
+					if state.log_max_size_bytes > 0
+						&& state.bytes_written + log_bytes > state.log_max_size_bytes
+					{
+						needs_rotation = true;
+					} else if state.log_rotation_interval_secs > 0 {
+						if let Ok(age) = SystemTime::now().duration_since(state.created_at) {
```
benthecarman: We're already getting the system time in `format_timestamp` for the log line, we don't need to fetch it again. Could clean up by calling `SystemTime::now()` at the top of this function and then passing it into `format_timestamp`.

benthecarman (changes requested):

ldk-server/src/util/logger.rs
Anyitechs: Done. Thank you

Anyitechs:
Addressed latest review comments and took the opportunity to squash fixup commits. 

Ready for another look!

ldk-server/src/util/logger.rs
```diff
@@ -8,74 +8,150 @@
...
-/// - A file specified during initialization
+/// - A file specified during initialization (if enabled)
 ///
 /// All log messages follow the format:
 /// `[TIMESTAMP LEVEL TARGET FILE:LINE] MESSAGE`
 ///
 /// Example: `[2025-12-04T10:30:45Z INFO ldk_server:42] Starting up...`
 ///
-/// The logger handles SIGHUP for log rotation by reopening the file handle when signaled.
+/// The logger does a native size/time-based rotation and retains the last 5 logs by default, if `max_rotated_files` is unset.
```
benthecarman: `log_max_files` not `max_rotated_files`

ldk-server/src/util/config.rs
```diff
@@ -257,6 +265,22 @@ impl ConfigBuilder {
...
+			self.log_rotation_interval_hours = Some(log_rotation_interval_hours);
+		}
+
+		if let Some(log_max_files) = args.log_max_files {
+			self.log_max_files = Some(log_max_files);
+		}
+
+		if let Some(log_file_path) = args.log_to_file {
+			self.log_to_file = Some(log_file_path);
+		}
```
benthecarman:
```suggestion
		if let Some(log_to_file) = args.log_to_file {
			self.log_to_file = Some(log_to_file);
		}
```

contrib/ldk-server-config.toml
```diff
@@ -16,11 +15,16 @@ dir_path = "/tmp/ldk-server/"                 # Path for LDK and BDK data persis
 [log]
 level = "Debug"                               # Log level (Error, Warn, Info, Debug, Trace)
 #file = "/tmp/ldk-server/ldk-server.log"      # Log file path
+# Enabling `log_to_file` automatically rotate logs at `max_size_mb` or `rotation_interval_hours`.
```
benthecarman:
```suggestion
# Enabling `log_to_file` automatically rotates logs at `max_size_mb` or `rotation_interval_hours`. 
```

ldk-server/src/util/logger.rs
```diff
@@ -8,74 +8,150 @@
...
+			return Err(io::Error::other("Logger state mutex poisoned"));
 		}
+		Ok(())
+	}
+
+	/// Flushes the current file, renames it with a timestamp, opens a fresh log,
+	/// and synchronously deletes older log files.
+	fn rotate(&self, state: &mut LoggerState) -> Result<(), io::Error> {
+		state.file.flush()?;
+		let log_file_path = self.log_file_path.as_ref().unwrap();
```
benthecarman: just to be safe, can we just return `Ok(()` if `log_file_path` is `None`. If we aren't logging to file then rotating is meaning less.

ldk-server/src/util/logger.rs
```diff
@@ -8,74 +8,150 @@
...
+		let rotated_path = PathBuf::from(new_path);
+
+		fs::rename(log_file_path, &rotated_path)?;
+
+		let new_file = open_log_file(log_file_path)?;
+		state.file = LineWriter::new(new_file);
+
+		// Reset our rotation triggers for the new file
+		state.bytes_written = 0;
+		state.created_at = SystemTime::now();
```
benthecarman: we can also use the `now` we call in log here

ldk-server/src/util/logger.rs
```diff
@@ -158,6 +248,45 @@ fn open_log_file(log_file_path: &Path) -> Result<File, io::Error> {
...
+						// The timestamp format "YYYY-MM-DDTHH-MM-SSZ" is 20 chars, contains 'T', and ends with 'Z'
+						suffix.len() == 20 && suffix.contains('T') && suffix.ends_with('Z')
+					})
+				})
+				.unwrap_or(false)
+		})
+		.collect();
+
+	// Sort by modification time (oldest first)
+	entries.sort_by_key(|e| e.metadata().and_then(|m| m.modified()).unwrap_or(SystemTime::now()));
```
benthecarman: should use `sort_by_cached_key` here

ldk-server/src/util/logger.rs
```diff
@@ -8,74 +8,150 @@
...
+
+		let new_file = open_log_file(log_file_path)?;
+		state.file = LineWriter::new(new_file);
+
+		// Reset our rotation triggers for the new file
+		state.bytes_written = 0;
+		state.created_at = SystemTime::now();
+
+		// Clean up old log files
+		if let Err(e) = cleanup_old_logs(log_file_path, state.log_max_files) {
```
benthecarman: We are doing this cleanup logs while holding the lock on the `LoggerState`. This means we are blocking all logs until this is done. We could put this after we drop the `LoggerState` in `log()` to prevent that.

ldk-server/src/util/logger.rs
```diff
@@ -8,74 +8,150 @@
...
+			eprintln!("Failed to clean up old log files: {}", e);
```
benthecarman: should use `log::error!` here

benthecarman (changes requested):
few more small things and this should be good!

Anyitechs:
Please excuse the delay here! 

I've addressed all the latest comments here [c8386a3](https://github.com/lightningdevkit/ldk-server/pull/189/commits/c8386a3bd3aa0e1dbfdc10c153c87fd0015ca5fc). Let me know if/when I can squash.

ldk-server/src/util/logger.rs
Anyitechs: Done. Thanks

ldk-server/src/util/logger.rs
```diff
@@ -86,62 +163,89 @@ impl Log for ServerLogger {
...
+						if let Ok(age) = now.duration_since(state.created_at) {
+							if age.as_secs() > state.log_rotation_interval_secs {
+								needs_rotation = true;
+							}
+						}
+					}
+
+					if needs_rotation {
+						if let Err(e) = self.rotate(&mut state, now) {
+							error!("Failed to rotate log file: {e}");
```
benthecarman: sorry I led you astray earlier, this will actually cause a dead lock because we have the mutex of the logger state now but will need it again for `error!` so we need this to be a `eprintln`

contrib/ldk-server-config.toml
```diff
@@ -16,11 +15,16 @@ dir_path = "/tmp/ldk-server/"                 # Path for LDK and BDK data persis
 [log]
 level = "Debug"                               # Log level (Error, Warn, Info, Debug, Trace)
 #file = "/tmp/ldk-server/ldk-server.log"      # Log file path
+# Enabling `log_to_file` automatically rotates logs at `max_size_mb` or `rotation_interval_hours`.
```
benthecarman: Can we be a little more clear here? We say 

> Enabling `log_to_file` automatically rotates logs at ...

which makes it sound like `log_to_file` enables rotation but it actually enables logging to the file, the rotation is the other config options. 

maybe something like:

> Enabling `log_to_file` writes logs to the configured file while still keeping the stdin/stdout logs available too. Logs files are automatically rotated at `max_size_mb` or `rotation_interval_hours`. To disable ...

docs/configuration.md
```diff
@@ -66,8 +61,12 @@ Where persistent data is stored. Defaults to `~/.ldk-server/` on Linux and
 
 ### `[log]`
 
-Log level and file path. The server reopens the log file on `SIGHUP`, which integrates with
-standard `logrotate` setups.
+Controls logging behavior. By default, `log_to_file` is `true` and logs are also written 
+to `stdout`/`stderr`.
+
+If `log_to_file` is enabled, the server performs internal rotation and retention
```
benthecarman: Same here, can copy the sentence structure above

docs/operations.md
```diff
@@ -21,13 +21,18 @@ The server handles `SIGTERM` and `CTRL-C` (SIGINT). On receipt, it:
...
-> configured, the log file will grow indefinitely and can eventually fill your disk. A full
-> disk can prevent the node from persisting channel state, risking fund loss.
+By default, LDK Server logs to `stdout`/`stderr` and also to file. When running under `systemd` or Docker, 
+this allows the environment (e.g., `journald`) to handle persistence, rotation, and 
+compression automatically.
 
-The server reopens its log file on `SIGHUP`. This integrates with standard `logrotate`. Save
-the following config to `/etc/logrotate.d/ldk-server` (adjust the log path to match your
-setup):
+If you enable `log_to_file` in the configuration, LDK Server will automatically rotate
```
benthecarman: again here

ldk-server/src/util/logger.rs
Anyitechs: Done. Thanks!

contrib/ldk-server-config.toml
Anyitechs: Done

Anyitechs:
Addressed latest review comments and squashed fixups. 

Ready for another look!

docs/operations.md
```diff
@@ -21,13 +21,18 @@ The server handles `SIGTERM` and `CTRL-C` (SIGINT). On receipt, it:
...
-the following config to `/etc/logrotate.d/ldk-server` (adjust the log path to match your
-setup):
+If you enable `log_to_file` in the configuration, LDK Server writes logs to the configured file while still 
+keeping the stdin/stdout logs available too. Logs files are automatically rotated at `max_size_mb` or `rotation_interval_hours`, and the last `max_files` uncompressed log files are retained. But you can disable 
+the internal rotation and keep logging to file by setting the `max_size_mb` and `rotation_interval_hours` 
+params to `0`.
+
+If you prefer to use system `logrotate` for file logs, the server still reopens its log 
+file on `SIGHUP`. Save the following config to `/etc/logrotate.d/ldk-server` 
+(adjust the log path to match your setup):
```
benthecarman: Can we fix the formatting on this file. The lines are all very different lengths

contrib/ldk-server-config.toml
```diff
@@ -16,11 +15,18 @@ dir_path = "/tmp/ldk-server/"                 # Path for LDK and BDK data persis
 [log]
 level = "Debug"                               # Log level (Error, Warn, Info, Debug, Trace)
 #file = "/tmp/ldk-server/ldk-server.log"      # Log file path
+# Enabling `log_to_file` writes logs to the configured file while still keeping 
+# the stdin/stdout logs available too. Logs files are automatically rotated at # `max_size_mb` or `rotation_interval_hours`.
```
benthecarman: `stdout/stderr` not `stdin/stdout`

benthecarman (commented):
couple docs fixes then we're good

benthecarman (approved):

PR #192, opened 2026-04-16
Bump LDK Node dependency to `c754e2fe85c70741b5e370334cd16856c615265e`

tankyleo:
Also make the LSPS2 configuration setting consistent between LDK Server and LDK Node.


tankyleo:
A follow-up to https://github.com/lightningdevkit/ldk-node/pull/873

benthecarman (approved):

PR #193, opened 2026-04-17
feat: channel events

### LINKED ISSUE #134: Add Channel state change event publishing to grpc subscriptions

frnandu:
fixes #134

### DIFF
--- a/e2e-tests/tests/e2e.rs
+++ b/e2e-tests/tests/e2e.rs
@@ -12,7 +12,8 @@ use std::time::Duration;
 
 use e2e_tests::{
 	find_available_port, mine_and_sync, run_cli, run_cli_raw, setup_funded_channel,
-	wait_for_onchain_balance, LdkServerConfig, LdkServerHandle, TestBitcoind,
+	wait_for_onchain_balance, wait_for_usable_channel, LdkServerConfig, LdkServerHandle,
+	TestBitcoind,
 };
 use hex_conservative::{DisplayHex, FromHex};
 use ldk_node::bitcoin::hashes::{sha256, Hash};
@@ -21,10 +22,10 @@ use ldk_node::lightning::offers::offer::Offer;
 use ldk_node::lightning_invoice::Bolt11Invoice;
 use ldk_server_client::client::EventStream;
 use ldk_server_client::ldk_server_grpc::api::{
-	Bolt11ReceiveRequest, Bolt12ReceiveRequest, OnchainReceiveRequest,
+	Bolt11ReceiveRequest, Bolt12ReceiveRequest, OnchainReceiveRequest, OpenChannelRequest,
 };
 use ldk_server_client::ldk_server_grpc::events::event_envelope::Event;
-use ldk_server_client::ldk_server_grpc::events::EventEnvelope;
+use ldk_server_client::ldk_server_grpc::events::{ChannelState, EventEnvelope};
 use ldk_server_client::ldk_server_grpc::types::{
 	bolt11_invoice_description, Bolt11InvoiceDescription,
 };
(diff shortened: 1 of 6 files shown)


Anyitechs (commented):
Thank you for looking into this. Please fix CI if this is ready for review (The proto files are now out of date and needs to be updated, and the changes introduced needs to be formatted too)

ldk-server/src/main.rs
```diff
@@ -12,6 +12,7 @@ mod io;
 mod service;
 mod util;
 
+use std::collections::HashSet;
```
Anyitechs, line 15: Do we want to use the `HashSet` from `HashBrown` here instead, since it's already in the tree?

ldk-server/src/main.rs
```diff
@@ -555,6 +634,131 @@ fn send_event_and_upsert_payment(
...
+
+fn is_channel_open_failure(reason: Option<&ClosureReason>) -> bool {
+	matches!(
+		reason,
+		Some(ClosureReason::FundingTimedOut)
+			| Some(ClosureReason::DisconnectedPeer)
+			| Some(ClosureReason::CounterpartyCoopClosedUnfundedChannel)
+			| Some(ClosureReason::LocallyCoopClosedUnfundedChannel)
+			| Some(ClosureReason::FundingBatchClosure)
+	)
```
benthecarman: instead of like this, can we do a `match reason { ...` then we can explicitly set true or false for each and we'll get a compile error when new reasons are added in the future

ldk-server/src/main.rs
```diff
@@ -555,6 +634,131 @@ fn send_event_and_upsert_payment(
...
+	match reason {
+		Some(ClosureReason::HolderForceClosed { .. })
+		| Some(ClosureReason::LocallyInitiatedCooperativeClosure)
+		| Some(ClosureReason::LocallyCoopClosedUnfundedChannel) => events::ChannelClosureInitiator::Local,
+		Some(ClosureReason::CounterpartyForceClosed { .. })
+		| Some(ClosureReason::CounterpartyInitiatedCooperativeClosure)
+		| Some(ClosureReason::CounterpartyCoopClosedUnfundedChannel) => {
+			events::ChannelClosureInitiator::Remote
+		},
+		Some(_) => events::ChannelClosureInitiator::Unknown,
```
benthecarman: can we enumerate all the other ones still instead of just `Some(_)` for same reason

ldk-server/src/main.rs
```diff
@@ -555,6 +634,131 @@ fn send_event_and_upsert_payment(
...
+			}))
+		},
+		ClosureReason::PeerFeerateTooLow {
+			peer_feerate_sat_per_kw,
+			required_feerate_sat_per_kw,
+		} => Some(Details::PeerFeerateTooLow(events::PeerFeerateTooLowDetails {
+			peer_feerate_sat_per_kw: *peer_feerate_sat_per_kw,
+			required_feerate_sat_per_kw: *required_feerate_sat_per_kw,
+		})),
+		_ => None,
```
benthecarman: again, can we enumerate them all

e2e-tests/tests/e2e.rs
```diff
@@ -410,6 +411,160 @@ async fn test_cli_open_channel() {
 	assert!(!output["user_channel_id"].as_str().unwrap().is_empty());
 }
 
+#[tokio::test]
+async fn test_subscribe_events_channel_state_lifecycle_pending_ready_closed() {
```
benthecarman, line 417: for both of these tests, can we check the events on both sides?

benthecarman: also we should be able to assert more things on these events

benthecarman (changes requested):

ldk-server/src/main.rs
frnandu: switching would add a new explicit dependency for little benefit in this case. `std::collections::HashSet` is the idiomatic default and already optimized AFAIK

benthecarman:
something weird happened when you merged main, can you just rebase your commit on main?

benthecarman:
commit structure is still weird here. should just be the one for the feature

ldk-server/src/main.rs
```diff
@@ -272,6 +274,11 @@ fn main() {
 			}
 		};
 		let event_node = Arc::clone(&node);
+		let mut ready_channel_ids: HashSet<String> = event_node
+			.list_channels()
+			.into_iter()
+			.map(|channel| channel.channel_id.0.to_lower_hex_string())
```
benthecarman: This list includes pending channels, need to filter by `.is_channel_ready`.

Also why use the string here, why not just the `ChannelId` type to be more safe

ldk-server/src/main.rs
```diff
@@ -324,38 +333,110 @@ fn main() {
...
+								counterparty_node_id,
+								funding_txo,
+							} => {
+								info!(
+									"CHANNEL_READY: {} from counterparty {:?}",
+									channel_id, counterparty_node_id
+								);
+
+								let channel_id_hex = channel_id.0.to_lower_hex_string();
+								ready_channel_ids.insert(channel_id.clone());
```
benthecarman: ci failing because un needed clone

benthecarman (commented):
lgtm once you fix clippy issue and squash commits

benthecarman (approved):

PR #194, opened 2026-04-17
`ldk-server-client`: Add `uniffi` bindings

tnull:
Not quite sure we want to add this at this point, but I needed it for a personal project, so vibecoded it. Draft for now.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -51,6 +51,48 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7c02d123df017efcdfbd739ef81735b36c5ba83ec3c59c80a9d7ecc718f92e50"
 
 [[package]]
+name = "askama"
+version = "0.13.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


benthecarman:
idk if this really makes sense, grpc already kinda gives us this for free. Main benefit would be not having to implement auth but that shouldn't be very hard

tnull:
> idk if this really makes sense, grpc already kinda gives us this for free. Main benefit would be not having to implement auth but that shouldn't be very hard

Hmm, well, even the current very basic scheme kept me (and my Claude instance at the time) from re-implementing the client-side logic in Kotlin, and I assume that will just get more of a lift when we switch to anything beyond (macaroons, etc). Yes, gRPC/protobuf is a pretty open format with good platform support, but IMO offering bindings would make it much more simple for people to integrate some kind of client-side code. Note that with `uniffi`, we get Swift, Kotlin, Pyhon, C#, and notably Go for free. So, just as an example, if LND ever wanted to add integration tests with LDK Server they could simply use the client-side Go bindings in their test harness.

Note that (in contrast to LDK Node's bindings) I spent no time to even attempt to DRY them up/reduce the changeset here. So I expect the diff could be quite a bit smaller if we tried.

PR #195, opened 2026-04-21
Configurable `HumanReadableName` resolver

tnull:
Bump LDK Node and add configurable HRN resolution.

### DIFF
--- a/ldk-server/src/util/config.rs
+++ b/ldk-server/src/util/config.rs
@@ -15,6 +15,7 @@ use std::{fs, io};
 use clap::Parser;
 use ldk_node::bitcoin::secp256k1::PublicKey;
 use ldk_node::bitcoin::Network;
+use ldk_node::config::{HRNResolverConfig, HumanReadableNamesConfig};
 use ldk_node::lightning::ln::msgs::SocketAddress;
 use ldk_node::lightning::routing::gossip::NodeAlias;
 use ldk_node::liquidity::LSPS2ServiceConfig;
@@ -61,6 +62,7 @@ pub struct Config {
 	pub metrics_username: Option<String>,
 	pub metrics_password: Option<String>,
 	pub tor_config: Option<TorConfig>,
+	pub hrn_config: HumanReadableNamesConfig,
 }
 
 #[derive(Debug, Clone, PartialEq, Eq)]
@@ -114,6 +116,7 @@ struct ConfigBuilder {
 	metrics_username: Option<String>,
 	metrics_password: Option<String>,
 	tor_proxy_address: Option<String>,
+	hrn: Option<HrnTomlConfig>,
 }
 
 impl ConfigBuilder {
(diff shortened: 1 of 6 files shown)


ldk-server/src/util/config.rs
```diff
@@ -505,6 +519,52 @@ struct TomlTorConfig {
...
+}
+
+impl TryFrom<HrnTomlConfig> for HumanReadableNamesConfig {
+	type Error = io::Error;
+
+	fn try_from(value: HrnTomlConfig) -> Result<Self, Self::Error> {
+		let mut config = HumanReadableNamesConfig::default();
+
+		match value.mode.as_deref() {
+			None | Some("dns") => {},
```
benthecarman: here we assume that the default is `HRNResolverConfig::Dns` and then modify the settings below. Would be safer to have the logic of the `if let HRNResolverConfig::Dns ...` in this arm so if the default ever changed (say to http to support lnurl) then we could better handle it and not silently drop the `dns_server_address` and `enable_resolution_service` settings.

ldk-server/src/util/config.rs
```diff
@@ -505,6 +519,52 @@ struct TomlTorConfig {
...
+			Some("blip32") => {
+				config.resolution_config = HRNResolverConfig::Blip32;
```
benthecarman: Would be good to throw an error if `enable_resolution_service` or `dns_server_address` is set. Otherwise users could be confused that they have something even though we are silently dropping the config option

ldk-server/src/util/config.rs
```diff
@@ -505,6 +519,52 @@ struct TomlTorConfig {
...
+					format!("Invalid HRN mode '{}' configured; expected 'dns' or 'blip32'", other),
+				))
+			},
+		}
+
+		if let HRNResolverConfig::Dns { dns_server_address, enable_hrn_resolution_service } =
+			&mut config.resolution_config
+		{
+			if let Some(addr) = value.dns_server_address.as_deref() {
+				*dns_server_address = SocketAddress::from_str(addr).map_err(|e| {
```
benthecarman: Can we do something here to assume port 53 if its not provided?

benthecarman (changes requested):
would be nice to one day add http support as well but mostly looks good!

ldk-server/src/util/config.rs
tnull: Done.

ldk-server/src/util/config.rs
tnull: Done.

ldk-server/src/util/config.rs
tnull: Done.

tnull:
Addressed pending feedback.

benthecarman (commented):
lgtm on squash

tnull:
> lgtm on squash

Anyitechs (approved):
Looks good, thanks!

benthecarman (approved):

PR #196, opened 2026-04-21
Add `--rescan-from-height` CLI argument

tnull:
Based on https://github.com/lightningdevkit/ldk-node/pull/884

We expose `--rescan-from-height` and `--force-wallet-full-scan` CLI arguments. First we bump the LDK Node dependency though, to include the changes from #884.

### DIFF
--- a/ldk-server/src/util/config.rs
+++ b/ldk-server/src/util/config.rs
@@ -702,6 +702,13 @@ pub struct ArgsConfig {
 		help = "Tor daemon SOCKS proxy address. Only connections to OnionV3 peers will be made via this proxy; other connections (IPv4 peers, Electrum server) will not be routed over Tor."
 	)]
 	tor_proxy_address: Option<String>,
+
+	#[arg(
+		long,
+		env = "LDK_SERVER_RESCAN_FROM_HEIGHT",
+		help = "Triggers wallet recovery on the next startup. On Bitcoin Core RPC/REST chain sources, the given block height is used as the wallet birthday so the chain is rescanned from that height (useful on pruned nodes when the wallet's birthday is known). On Esplora/Electrum chain sources the given height is ignored, but setting the flag still forces a one-shot BDK `full_scan` so funds sent to previously-unknown addresses are re-discovered. Only takes effect on first startup (i.e. for a freshly-persisted wallet)."
+	)]
+	pub rescan_from_height: Option<u32>,
 }
 
 pub fn load_config(args: &ArgsConfig) -> io::Result<Config> {
--- a/ldk-server/src/util/config.rs
+++ b/ldk-server/src/util/config.rs
@@ -98,9 +98,11 @@ pub enum ChainSource {
 	},
 	Electrum {
 		server_url: String,
+		force_wallet_full_scan: bool,
 	},
 	Esplora {
 		server_url: String,
+		force_wallet_full_scan: bool,
 	},
 }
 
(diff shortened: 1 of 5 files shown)


ldk-server/src/util/config.rs
```diff
@@ -363,6 +385,13 @@ impl ConfigBuilder {
 		}
 
 		let chain_source = if rpc_configured {
+			if self.force_wallet_full_scan {
+				return Err(io::Error::new(
+					io::ErrorKind::InvalidInput,
+					"`--force-wallet-full-scan` requires the Electrum or Esplora chain source.",
```
benthecarman, line 391: we should reference `rescan-from-height` in this error message

ldk-server/src/util/config.rs
```diff
@@ -375,11 +404,35 @@ impl ConfigBuilder {
...
+				rpc_user,
+				rpc_password,
+				wallet_rescan_from_height: self.rescan_from_height,
+			}
 		} else if let Some(url) = self.electrum_url {
-			ChainSource::Electrum { server_url: url }
+			if self.rescan_from_height.is_some() {
+				return Err(io::Error::new(
+					io::ErrorKind::InvalidInput,
+					"`--rescan-from-height` requires the bitcoind RPC chain source.",
```
benthecarman, line 418: we should reference `--force-wallet-full-scan` in this error message

ldk-server/src/util/config.rs
```diff
@@ -375,11 +404,35 @@ impl ConfigBuilder {
...
+			ChainSource::Electrum {
+				server_url: url,
+				force_wallet_full_scan: self.force_wallet_full_scan,
+			}
 		} else if let Some(url) = self.esplora_url {
-			ChainSource::Esplora { server_url: url }
+			if self.rescan_from_height.is_some() {
+				return Err(io::Error::new(
+					io::ErrorKind::InvalidInput,
+					"`--rescan-from-height` requires the bitcoind RPC chain source.",
```
benthecarman, line 429: same here

benthecarman (changes requested):
I see you excluded these from the toml config, makes sense because its not really something you set persistently, just for a single run. Should we exclude the env variations for these too then? Env variables feel similar to me but maybe that's just my systemd preference coming out

ldk-server/src/util/config.rs
tnull: Hmm, could you expand what you're thinking of there? Seems they are slightly orthogonal, so not sure why we need the cross-referencing?

benthecarman: If someone is setting `--force-wallet-full-scan` with a bitcoind backend I presume they are trying to rescan for missing funds and we should direct them to `--rescan-from-height` which is the config for their backend

tnull: But they do slightly different things? Isn't it even more confusing to point to something that works differently than the thing you're trying to do?  Note we already do a full scan on first startup anyways, so differntly to `--rescan-from-height` it's not required to import an old wallet from seed necessarily.

benthecarman: I think the "works differently" here is mostly just an implementation detail to user that isn't really relevant to the outcome they are looking for.

Don't feel too strongly though, so if you dont want to, that's fine

tnull: > I think the "works differently" here is mostly just an implementation detail to user that isn't really relevant to the outcome they are looking for.

Hmm, but the usecase is different to begin with? `--rescan-from-height` should be used once when you import from pre-existing seed, while `--force-full-scan` is more a utility that you only need if for some reason you discover there are still funds missing after the fact?

benthecarman: They are both "I'm missing funds, rescan chain" but yeah, I don't feel too strongly either way

benthecarman (approved):

PR #197, opened 2026-04-21
Make entropy source configurable, default to mnemonic

tnull:
We switch to, by default, creating and reading a BIP39 mnemonic file rather than raw bytes. We however detect the old behavior and will keep using `keys_seed` if it is present.

### DIFF
--- a/ldk-server/src/util/config.rs
+++ b/ldk-server/src/util/config.rs
@@ -61,6 +61,18 @@ pub struct Config {
 	pub metrics_username: Option<String>,
 	pub metrics_password: Option<String>,
 	pub tor_config: Option<TorConfig>,
+	pub entropy: EntropyConfig,
+}
+
+/// Configuration for the node's entropy source.
+///
+/// When both `mnemonic_file` and `seed_file` are unset, the node defaults to loading or
+/// generating a BIP39 mnemonic at `<storage_dir>/keys_mnemonic`. If a legacy raw-seed file
+/// exists at `<storage_dir>/keys_seed`, it is used for backwards compatibility.
+#[derive(Debug, Default, Clone, PartialEq, Eq)]
+pub struct EntropyConfig {
+	pub mnemonic_file: Option<String>,
+	pub seed_file: Option<String>,
 }
 
 #[derive(Debug, Clone, PartialEq, Eq)]
@@ -114,6 +126,7 @@ struct ConfigBuilder {
 	metrics_username: Option<String>,
 	metrics_password: Option<String>,
 	tor_proxy_address: Option<String>,
+	entropy: EntropyConfig,
 }
 
 impl ConfigBuilder {
@@ -130,6 +143,11 @@ impl ConfigBuilder {
 			self.pathfinding_scores_source_url =
 				node.pathfinding_scores_source_url.or(self.pathfinding_scores_source_url.clone());
 			self.rgs_server_url = node.rgs_server_url.or(self.rgs_server_url.clone());
+			if let Some(entropy) = node.entropy {
+				self.entropy.mnemonic_file =
+					entropy.mnemonic_file.or(self.entropy.mnemonic_file.take());
+				self.entropy.seed_file = entropy.seed_file.or(self.entropy.seed_file.take());
+			}
 		}
 
 		if let Some(storage) = toml.storage {
@@ -402,6 +420,13 @@ impl ConfigBuilder {
 			})
 			.transpose()?;
 
+		if self.entropy.mnemonic_file.is_some() && self.entropy.seed_file.is_some() {
+			return Err(io::Error::new(
+				io::ErrorKind::InvalidInput,
+				"Only one of `node.entropy.mnemonic_file` and `node.entropy.seed_file` may be configured.".to_string(),
+			));
+		}
+
 		Ok(Config {
 			network,
 			listening_addrs,
@@ -422,6 +447,7 @@ impl ConfigBuilder {
 			metrics_username,
 			metrics_password,
 			tor_config: tor_proxy_address.map(|proxy_address| TorConfig { proxy_address }),
+			entropy: self.entropy,
 		})
 	}
 }
@@ -450,6 +476,13 @@ struct NodeConfig {
 	alias: Option<String>,
 	pathfinding_scores_source_url: Option<String>,
 	rgs_server_url: Option<String>,
+	entropy: Option<NodeEntropyTomlConfig>,
+}
+
+#[derive(Deserialize, Serialize)]
+struct NodeEntropyTomlConfig {
+	mnemonic_file: Option<String>,
+	seed_file: Option<String>,
 }
 
 #[derive(Deserialize, Serialize)]
@@ -936,6 +969,7 @@ mod tests {
 			tor_config: Some(TorConfig {
 				proxy_address: SocketAddress::from_str("127.0.0.1:9050").unwrap(),
 			}),
+			entropy: EntropyConfig::default(),
 		};
 
 		assert_eq!(config.listening_addrs, expected.listening_addrs);
@@ -1241,6 +1275,7 @@ mod tests {
 			metrics_username: None,
 			metrics_password: None,
 			tor_config: None,
+			entropy: EntropyConfig::default(),
 		};
 
 		assert_eq!(config.listening_addrs, expected.listening_addrs);
@@ -1350,6 +1385,7 @@ mod tests {
 			tor_config: Some(TorConfig {
 				proxy_address: SocketAddress::from_str("127.0.0.1:9050").unwrap(),
 			}),
+			entropy: EntropyConfig::default(),
 		};
 
 		assert_eq!(config.listening_addrs, expected.listening_addrs);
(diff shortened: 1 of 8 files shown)


benthecarman:
concept ack

tnull:
Should be good for review.

ldk-server/src/util/config.rs
```diff
@@ -137,6 +150,11 @@ impl ConfigBuilder {
 			self.async_payments_role =
 				node.async_payments_role.or(self.async_payments_role.clone());
 			self.rgs_server_url = node.rgs_server_url.or(self.rgs_server_url.clone());
+			if let Some(entropy) = node.entropy {
+				self.entropy.mnemonic_file =
```
Anyitechs: We use `take()` here and below because we immediately want to wipe the `mnemonic_file`/`seed_file` from memory after parsing, correct?

Anyitechs (commented):
Thanks! 

Just curious, is there a reason the entropy source are only configurable via the toml file and not added to the env var as well?

ldk-server/src/util/entropy.rs
```diff
@@ -0,0 +1,204 @@
...
+
+	let legacy_seed_path = storage_dir.join(LEGACY_SEED_FILE);
+	if entropy_config.mnemonic_file.is_none() && legacy_seed_path.exists() {
+		info!(
+			"Detected legacy raw seed file at {}; continuing to use it. New installs use a BIP39 mnemonic by default.",
+			legacy_seed_path.display()
+		);
+		return NodeEntropy::from_seed_path(legacy_seed_path.to_string_lossy().into_owned())
+			.map_err(|e| io::Error::new(io::ErrorKind::InvalidData, e.to_string()));
+	}
```
benthecarman: Add handling for the like 5 of us seems a bit weird. I'd rather just have us set the config for the legacy path and then not have to handle this in the code.

ldk-server/src/util/config.rs
```diff
@@ -64,6 +64,18 @@ pub struct Config {
...
+
+/// Configuration for the node's entropy source.
+///
+/// When both `mnemonic_file` and `seed_file` are unset, the node defaults to loading or
+/// generating a BIP39 mnemonic at `<storage_dir>/keys_mnemonic`. If a legacy raw-seed file
+/// exists at `<storage_dir>/keys_seed`, it is used for backwards compatibility.
+#[derive(Debug, Default, Clone, PartialEq, Eq)]
+pub struct EntropyConfig {
+	pub mnemonic_file: Option<String>,
+	pub seed_file: Option<String>,
```
benthecarman: since we disallow both being `Some` this would work better as an enum imo

```rust
pub enum EntropyConfig {
  Mnemonic(String),
  SeedFile(String),
  Default, // default mnemonic path in storage dir
}
```

ldk-server/src/util/entropy.rs
```diff
@@ -0,0 +1,204 @@
...
+	};
+
+	Ok(NodeEntropy::from_bip39_mnemonic(mnemonic, None))
+}
+
+fn write_mnemonic_file(path: &Path, mnemonic: &Mnemonic) -> io::Result<()> {
+	let mut f = fs::OpenOptions::new().create_new(true).write(true).mode(0o600).open(path)?;
+	writeln!(f, "{}", mnemonic)?;
+	f.sync_all()?;
+	fs::set_permissions(path, fs::Permissions::from_mode(0o600))?;
```
benthecarman: shouldn't the `f.sync_all()` be after we set permissions?

benthecarman (changes requested):

ldk-server/src/util/config.rs
tnull: Yeah, but now dropped.

ldk-server/src/util/entropy.rs
tnull: Addressed in a fixup.

ldk-server/src/util/config.rs
tnull: Done in a fixup.

ldk-server/src/util/entropy.rs
tnull: Done in a fixup.

ldk-server/src/util/entropy.rs
```diff
@@ -0,0 +1,185 @@
...
+			"Generated new BIP39 mnemonic at {}. Back up this file securely — it is required to recover on-chain funds.",
+			mnemonic_path.display()
+		);
+		mnemonic
+	};
+
+	Ok(NodeEntropy::from_bip39_mnemonic(mnemonic, None))
+}
+
+fn write_mnemonic_file(path: &Path, mnemonic: &Mnemonic) -> io::Result<()> {
```
benthecarman: can now use this function instead https://github.com/benthecarman/ldk-server/blob/a5d187e13d951dc09587a11a2e24ac889a3fe4b2/ldk-server/src/util/mod.rs#L22

ldk-server/src/util/entropy.rs
```diff
@@ -0,0 +1,185 @@
...
+
+use ldk_node::bip39::Mnemonic;
+use ldk_node::entropy::{generate_entropy_mnemonic, NodeEntropy};
+use log::info;
+
+use crate::util::config::EntropyConfig;
+
+const DEFAULT_MNEMONIC_FILE: &str = "keys_mnemonic";
+#[cfg(test)]
+const LEGACY_SEED_FILE: &str = "keys_seed";
```
benthecarman: can we move this just down into the tests below

benthecarman (changes requested):

lorenzolfm:
Hi @tnull. I think that this change doesn't do the 

> We however detect the old behavior and will keep using keys_seed if it is present.

Part you described.

I've did the following to test:

1. Ran ldk-server on main, let the software generate entropy for me.

```
 [INFO  ldk_node] Starting up LDK Node with node ID 0235b4adee490f4aba6bcb9dfb5d145d132625adb8229b08fea6bc295ef292577f on network: regtest
 [INFO  ldk_server] NODE_URI: 0235b4adee...77f@127.0.0.1:9735
 [INFO  ldk_server] gRPC service listening on 127.0.0.1:3536
 ```
 
2. Stopped the node, switched to this branch, build and ran. Without setting anything entropy related in the config file.

```
 [INFO  ldk_server::util::entropy] Generated new BIP39 mnemonic at /tmp/ldk-repro-pr197/ldk-storage/keys_mnemonic. Back up this file securely — it is required to recover on-chain funds.
 [ERROR ldk_node::builder] Failed to set up wallet: Descriptor mismatch for External keychain:
         loaded   wpkh([c0ff2044/84'/1'/0']tpubDCLz...)#7y546d6z 
         expected wpkh([2bd63f70/84'/1'/0']tpubDCDx...)#d5ncgc2j 
 [ERROR ldk_server] Failed to build LDK Node: Failed to setup onchain wallet.
  ```
  
It seems that:
 
 * existing `keys_seed` is ignored
 * A brand-new mnemonic is silently written to disk before any safety check runs
 * The node fails to start with a nasty error message
 
 Also, if I may, I'm not sure that the best path is ignoring **any** key, as it may confuse the user on what key he is actually using. Maybe the best path forward is fail to boot not only if both config options are present, but if the two files are present or something like that

tnull:
> Hi @tnull. I think that this change doesn't do the
> 
> > We however detect the old behavior and will keep using keys_seed if it is present.
> 
> Part you described.

Well, it used to, but Ben explicitly requested the backwards compatibility logic to be dropped in https://github.com/lightningdevkit/ldk-server/pull/197#discussion_r3268587555 as there are currently only ~5 nodes actually deployed for testing purposes and if this is the new default behavior prior to the actual release, we should be able to at some point even drop the legacy keys_seed behavior entirely.

ldk-server/src/util/entropy.rs
tnull: Done in 0cf92134eec53eeb35a31eb5e4e403368b6138a0

ldk-server/src/util/entropy.rs
tnull: Done in 0cf92134eec53eeb35a31eb5e4e403368b6138a0

benthecarman (commented):
Lgtm on squash

tankyleo:
> we should be able to at some point even drop the legacy keys_seed behavior entirely.

I'm happy to set the legacy keys seed explicitly, and would love to keep this behavior around long-term.

tnull:
> > we should be able to at some point even drop the legacy keys_seed behavior entirely.
> 
> I'm happy to set the legacy keys seed explicitly, and would love to keep this behavior around long-term.

Mhh, I mean its not much code, but more generally we're not providing compatibility guarantees prior to 0.1 final release, and the 3 of us running on mainnet can be expected to migrate once. Before we cut the release we should also take another good look at the codebase if there is any tech debt already that we want to remove while we're free to break API and backwards compat. (cc @benthecarman).

tankyleo:
> Mhh, I mean its not much code, but more generally we're not providing compatibility guarantees prior to 0.1 final release, and the 3 of us running on mainnet can be expected to migrate once.

I have no problem with required migrations prior to the release. Rather, I do not want to take a dependency on BIP39 to run ldk-server :)

tnull:
> I have no problem with required migrations prior to the release. Rather, I do not want to take a dependency on BIP39 to run ldk-server :)
 
Okay, would love to learn why that is, but let's take that discussion offline.

tnull:
> Lgtm on squash

benthecarman:
I think we discussed offline just removing the old file type completely. Do you still want to do that here?

tnull:
> I think we discussed offline just removing the old file type completely. Do you still want to do that here?

Done, dropped support for `keys_seed`.

benthecarman (approved):

PR #198, opened 2026-05-05
Add `network` field to `GetNodeInfoResponse`

randomlogin:
Expose the Bitcoin network the node is running on (e.g., "bitcoin", "testnet", "signet", "regtest") via the get-node-info API.

This was merged in https://github.com/lightningdevkit/ldk-node/pull/892

I've bumped the dependency revision of ldk-node to the latest one before breaking changes are introduced (https://github.com/lightningdevkit/ldk-node/pull/850).

### DIFF
--- a/ldk-server-grpc/src/proto/api.proto
+++ b/ldk-server-grpc/src/proto/api.proto
@@ -70,6 +70,9 @@ message GetNodeInfoResponse {
   // These are constructed from the announcement addresses and the node's public key.
   // Will be empty if no announcement addresses are configured.
   repeated string node_uris = 12;
+
+  // The Bitcoin network the node is running on (e.g., "bitcoin", "testnet", "signet", "regtest").
+  types.Network network = 13;
 }
 
 // Retrieve a new on-chain funding address.
--- a/ldk-server/Cargo.toml
+++ b/ldk-server/Cargo.toml
@@ -4,7 +4,7 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-ldk-node = { git = "https://github.com/lightningdevkit/ldk-node", rev = "21eea8c881790db7a90bcad4f7f45f341c72222b" }
+ldk-node = { git = "https://github.com/lightningdevkit/ldk-node", rev = "16eaa6f46308965f501904506be3ceecd293f358" }
 serde = { version = "1.0.203", default-features = false, features = ["derive"] }
 hyper = { version = "1", default-features = false, features = ["server", "http2"] }
 http-body-util = { version = "0.1", default-features = false }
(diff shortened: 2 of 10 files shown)


ldk-server-grpc/src/proto/api.proto
```diff
@@ -70,6 +70,9 @@ message GetNodeInfoResponse {
   // These are constructed from the announcement addresses and the node's public key.
   // Will be empty if no announcement addresses are configured.
   repeated string node_uris = 12;
+
+  // The Bitcoin network the node is running on (e.g., "bitcoin", "testnet", "signet", "regtest").
+  string network = 13;
```
benthecarman: instead of a string, can we make this a proper enum in the proto

benthecarman (changes requested):

ldk-server-grpc/src/proto/api.proto
randomlogin: Added an enum representing Network as well as `network_to_proto` function.

ldk-server/src/util/proto_adapter.rs
```diff
@@ -510,3 +511,14 @@ pub(crate) fn graph_node_to_proto(node: NodeInfo) -> ldk_server_grpc::types::Gra
...
 }
+
+pub(crate) fn network_to_proto(network: Network) -> ldk_server_grpc::types::Network {
+	use ldk_server_grpc::types::Network as ProtoNetwork;
+	match network {
+		Network::Bitcoin => ProtoNetwork::Bitcoin.into(),
+		Network::Testnet => ProtoNetwork::Testnet.into(),
+		Network::Testnet4 => ProtoNetwork::Testnet4.into(),
+		Network::Signet => ProtoNetwork::Signet.into(),
+		Network::Regtest => ProtoNetwork::Regtest.into(),
```
benthecarman: i don't think you need all these `.into()`s

ldk-server/Cargo.toml
```diff
@@ -4,7 +4,7 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-ldk-node = { git = "https://github.com/lightningdevkit/ldk-node", rev = "21eea8c881790db7a90bcad4f7f45f341c72222b" }
+ldk-node = { git = "https://github.com/lightningdevkit/ldk-node", rev = "16eaa6f46308965f501904506be3ceecd293f358" }
```
Anyitechs, line 7: Do we want to move this bump into it's own commit?

ldk-server/src/util/proto_adapter.rs
randomlogin: hmm yes; also spotted another problem, will patch promptly

ldk-server/Cargo.toml
randomlogin: Made it a distinct commit.

benthecarman (approved):

PR #200, opened 2026-05-05
Change enum serialization to lowercase

randomlogin:
Changed the serialization macros to lowercase enum names.

Strings like "COOPERATIVE_CLOSURE", "SUCCEEDED", or "SIGNET" look unnatural in CLI output and human-facing JSON, even though SCREAMING_SNAKE_CASE is the proto3 default. 

This relies on enum variants following the proto style guide (UPPER_SNAKE_CASE); a future variant using acronyms or mixed case (e.g. HTTPSEndpoint) would lowercase awkwardly. All current variants are clean.

Note: this is a breaking change for any existing JSON consumer of PaymentDirection, PaymentStatus, or BalanceSource.

### DIFF
--- a/ldk-server-grpc/src/serde_utils.rs
+++ b/ldk-server-grpc/src/serde_utils.rs
@@ -18,7 +18,7 @@ use std::fmt::Write;
 use serde::Serializer;
 
 /// Generates a serde serializer that converts an `i32` proto enum field to its
-/// string name via `from_i32()` and `as_str_name()`.
+/// lowercased string name via `from_i32()` and `as_str_name()`.
 macro_rules! stringify_enum_serializer {
 	($fn_name:ident, $enum_type:ty) => {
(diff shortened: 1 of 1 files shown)


benthecarman (commented):
tbh I don't really love this, seeing `status: SUCCEEDED` feels more like an enum whereas `status: succeeded` feels like someone hard coded a string to me

randomlogin:
> tbh I don't really love this, seeing `status: SUCCEEDED` feels more like an enum whereas `status: succeeded` feels like someone hard coded a string to me

Can it be so because you know the internals and the fact that there is an enum?
Funny enough, it makes me feel the contrary: in "naturally occurring" JSONs an uppercase field gives a suspicion that something went wrong in a piece of software and I received a response encoded for a machine and maybe I need to recheck if there was no error.

vincenzopalazzo:
I kind of tend to agree with @benthecarman

PR #207, opened 2026-05-12
Sign auth HMAC over request bodies

benthecarman:
Require authenticated gRPC requests to bind the HMAC to both the timestamp and the request body. This prevents a valid header from being replayed with different request contents during the allowed timestamp window.

Update the client and docs so callers generate signatures that match the new server contract.

Originally didn't do this to keep things simpler and we were exploring other auth options. Now that we've seemed to settle on this for now, may as well improve it.

### DIFF
--- a/ldk-server/src/service.rs
+++ b/ldk-server/src/service.rs
@@ -113,9 +113,16 @@ impl NodeService {
 // Maximum allowed time difference between client timestamp and server time (1 minute)
 const AUTH_TIMESTAMP_TOLERANCE_SECS: u64 = 60;
 
+fn compute_auth_hmac(api_key: &str, timestamp: u64, body: &[u8]) -> Hmac<sha256::Hash> {
+	let mut hmac_engine: HmacEngine<sha256::Hash> = HmacEngine::new(api_key.as_bytes());
+	hmac_engine.input(&timestamp.to_be_bytes());
+	hmac_engine.input(body);
+	Hmac::<sha256::Hash>::from_engine(hmac_engine)
+}
+
 /// Validates HMAC authentication from request headers.
-/// Uses timestamp-only HMAC (no body) since TLS guarantees integrity.
-fn validate_auth<B>(req: &Request<B>, api_key: &str) -> Result<(), LdkServerError> {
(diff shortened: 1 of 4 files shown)


ldk-server/src/service.rs
```diff
@@ -475,6 +499,20 @@ async fn handle_grpc_unary<
 	}
 }
 
+async fn read_request_body(body: Incoming) -> Result<bytes::Bytes, GrpcStatus> {
+	let limited_body = Limited::new(body, MAX_BODY_SIZE);
+	let bytes = match limited_body.collect().await {
+		Ok(collected) => collected.to_bytes(),
+		Err(_) => {
+			return Err(GrpcStatus::new(
```
lorenzolfm, line 507: maybe worthwhile logging instead of discarding the error?

ldk-server/src/service.rs
```diff
@@ -490,10 +528,8 @@ pub(crate) fn ldk_error_to_grpc_status(e: LdkServerError) -> GrpcStatus {
 mod tests {
 	use super::*;
 
-	fn compute_hmac(api_key: &str, timestamp: u64) -> String {
-		let mut hmac_engine: HmacEngine<sha256::Hash> = HmacEngine::new(api_key.as_bytes());
-		hmac_engine.input(&timestamp.to_be_bytes());
-		Hmac::<sha256::Hash>::from_engine(hmac_engine).to_string()
+	fn compute_hmac(api_key: &str, timestamp: u64, body: &[u8]) -> String {
+		compute_auth_hmac(api_key, timestamp, body).to_string()
```
lorenzolfm, line 532: useless indirection?

lorenzolfm (approved):
(not worth much)ACK 819bed81a3ddb38f5b84b4749c0426d72b064e86

ldk-server/src/service.rs
tnull: No, logging an error here pre-auth would introduce a potential DoS vector, as a malicious client could craft messages that fail to decode, hence spamming our logs.

tnull (approved):

PR #208, opened 2026-05-14
Expose async payments role config

benthecarman:
Allow operators to configure ldk-node's async payments role from the TOML file, CLI, or environment. This wires the resolved role into node construction and documents the client and server options.


lorenzolfm (commented):
ACK ddbb02d408a71a1da8fd437cbcd12400daabbe2c

joostjager (approved):

PR #211, opened 2026-05-17
fix: configuration parsing edge cases and startup panics

Ferryx349:
This PR addresses several edge cases in configuration parsing and resolves potential startup panics, improving the overall robustness of the node. It also cleans up a few minor code quality warnings flagged by modern `clippy`.

### DIFF
--- a/ldk-server/src/util/config.rs
+++ b/ldk-server/src/util/config.rs
@@ -387,6 +387,13 @@ impl ConfigBuilder {
 
 		let poll_metrics_interval = self.poll_metrics_interval;
 
+		if let Some(0) = poll_metrics_interval {
+			return Err(io::Error::new(
+				io::ErrorKind::InvalidInput,
+				"poll_metrics_interval must be greater than 0",
+			));
+		}
+
 		let metrics_username = self.metrics_username;
 		let metrics_password = self.metrics_password;
 
@@ -597,7 +604,11 @@ fn parse_dns_server_address(addr: &str) -> io::Result<SocketAddress> {
 	if let Ok(sa) = SocketAddress::from_str(addr) {
 		return Ok(sa);
 	}
-	let with_default_port = format!("{}:53", addr);
+	let with_default_port = if addr.contains(':') && !addr.starts_with('[') {
+		format!("[{}]:53", addr)
+	} else {
+		format!("{}:53", addr)
+	};
 	SocketAddress::from_str(&with_default_port).map_err(|e| {
 		io::Error::new(
 			io::ErrorKind::InvalidInput,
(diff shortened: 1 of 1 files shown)


ldk-server/src/util/config.rs
```diff
@@ -610,7 +601,11 @@ fn parse_dns_server_address(addr: &str) -> io::Result<SocketAddress> {
 	if let Ok(sa) = SocketAddress::from_str(addr) {
 		return Ok(sa);
 	}
-	let with_default_port = format!("{}:53", addr);
+	let with_default_port = if addr.contains(':') && !addr.starts_with('[') {
```
benthecarman, line 607: Can you add test cases for these

ldk-server-cli/src/types.rs
```diff
@@ -65,7 +65,7 @@ impl Amount {
 	///
 	/// Returns an error string if the value is not evenly divisible by 1000.
 	pub fn to_sat(self) -> Result<u64, String> {
-		if self.msats % 1000 != 0 {
+		if !self.msats.is_multiple_of(1000) {
```
benthecarman: Our msrv doesn't support this

benthecarman (changes requested):

ldk-server-cli/src/types.rs
Ferryx349: I was running a newer local toolchain for clippy and didn't realize that method was stabilized recently, will revert it back...

ldk-server/src/util/config.rs
Ferryx349: added test cases

ldk-server/src/main.rs
```diff
@@ -137,7 +137,7 @@ fn main() {
 		},
 	};
 
-	ldk_node_config.storage_dir_path = network_dir.to_str().unwrap().to_string();
+	ldk_node_config.storage_dir_path = network_dir.to_string_lossy().to_string();
```
benthecarman: I'd rather keep these panics than use `to_string_lossy` where we could be writing to a dir that the user doesn't expect

benthecarman (changes requested):
formatting is failing on ci as well

ldk-server/src/main.rs
Ferryx349: make sense, reverted it back and fixed formatting

ldk-server/src/util/config.rs
```diff
@@ -391,13 +383,17 @@ impl ConfigBuilder {
...
 		let pathfinding_scores_source_url = self.pathfinding_scores_source_url;
 
-		let async_payments_role =
-			self.async_payments_role.as_deref().map(parse_async_payments_role).transpose()?;
-
 		let metrics_enabled = self.metrics_enabled.unwrap_or(false);
 
 		let poll_metrics_interval = self.poll_metrics_interval;
 
+		if let Some(0) = poll_metrics_interval {
```
Anyitechs, line 390: No strong opinion here, but I actually don't see the value of this check. This already panics if the `poll_metrics_interval` is zero because the value is being passed as the period argument for the `tokio::time::interval()` function in `main`.

Ferryx349: my resoning was to fail early and gracefully, to return a config error upfront rather than letting tokio panic and dump a stack trace later in main....

however, i can revert it back if makes sense...

Anyitechs: > my resoning was to fail early and gracefully, to return a config error upfront rather than letting tokio panic and dump a stack trace later in main....

Oh, that works. Thank you!

benthecarman (approved):

PR #213, opened 2026-05-19
Expose introduction node direction in decoded offers

jkczyz:
A BOLT12 blinded path's introduction node may be identified by a directed short channel ID (SCID plus a direction byte selecting one of the channel's two endpoints), not just by node id. Previously the decoded offer dropped the direction byte, leaving callers unable to tell which endpoint was the introduction node. The two identification forms were also represented as independent optional fields, so the schema did not enforce that exactly one was set; mutual exclusivity relied on server convention.

Restructure `BlindedPath` so the introduction node is a `oneof` between a `node_id` string and a new `DirectedShortChannelId` message carrying both the SCID and a `ChannelDirection` enum. Populate it from the LDK `IntroductionNode::DirectedShortChannelId` variant.


benthecarman (approved):
thanks!

PR #215, opened 2026-05-20
Decode BOLT12 invoices in the DecodeInvoice API

vincenzopalazzo:
## Summary

- `DecodeInvoice` now also decodes a hex-encoded BOLT12 invoice, not just a BOLT11 invoice string.
- `DecodeInvoiceResponse` gains a `kind` field (`"bolt11"` or `"bolt12"`) identifying which invoice type was decoded.
- A BOLT12 invoice has no human-readable string form (it is exchanged as raw bytes over onion messages), so the input is expected to be hex-encoded and is parsed via LDK's `Bolt12Invoice::try_from`. Per the intentionally minimal scope, only `kind` is populated for a BOLT12 invoice; the remaining fields continue to apply to BOLT11 invoices.
- CLI / MCP / client docs updated to reflect the broadened input.

## Test plan

- [x] `cargo test` — new unit tests in `decode_invoice.rs` cover the BOLT12 invoice round-trip and rejection of unparseable / non-invoice input
- [x] e2e `test_cli_decode_invoice` and `test_mcp_live_tool_calls` assert `kind == "bolt11"`
- [x] `cargo check --all-features --all-targets`, `cargo fmt --all`, clippy clean for the changed code
- [x] Protobuf regenerated via `RUSTFLAGS="--cfg genproto" cargo build -p ldk-server-grpc`

AI tools: implemented with Claude Code (Claude Opus 4.7).

### DIFF
--- a/e2e-tests/tests/e2e.rs
+++ b/e2e-tests/tests/e2e.rs
@@ -193,6 +193,7 @@ async fn test_cli_decode_invoice() {
 	assert!(decoded["timestamp"].as_u64().unwrap() > 0);
 	assert!(decoded["min_final_cltv_expiry_delta"].as_u64().unwrap() > 0);
 	assert_eq!(decoded["is_expired"], false);
+	assert_eq!(decoded["kind"], "bolt11");
 
 	// Verify features — LDK BOLT11 invoices always set VariableLengthOnion, PaymentSecret,
 	// and BasicMPP.
--- a/e2e-tests/tests/mcp.rs
+++ b/e2e-tests/tests/mcp.rs
@@ -84,4 +84,5 @@ async fn test_mcp_live_tool_calls() {
 	assert_eq!(decode_invoice_json["destination"], server.node_id());
 	assert_eq!(decode_invoice_json["description"], "mcp decode");
 	assert_eq!(decode_invoice_json["amount_msat"], 50_000_000u64);
+	assert_eq!(decode_invoice_json["kind"], "bolt11");
 }
--- a/ldk-server-grpc/src/api.rs
+++ b/ldk-server-grpc/src/api.rs
@@ -1109,19 +1109,21 @@ pub struct GraphGetNodeResponse {
 	#[prost(message, optional, tag = "1")]
 	pub node: ::core::option::Option<super::types::GraphNode>,
 }
-/// Decode a BOLT11 invoice and return its parsed fields.
-/// This does not require a running node — it only parses the invoice string.
+/// Decode a BOLT11 or BOLT12 invoice and return its parsed fields.
+/// This does not require a running node — it only parses the invoice.
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[cfg_attr(feature = "serde", serde(rename_all = "snake_case"))]
 #[cfg_attr(feature = "serde", serde(default))]
 #[allow(clippy::derive_partial_eq_without_eq)]
 #[derive(Clone, PartialEq, ::prost::Message)]
 pub struct DecodeInvoiceRequest {
-	/// The BOLT11 invoice string to decode.
+	/// The invoice to decode: either a BOLT11 invoice string or a hex-encoded BOLT12 invoice.
 	#[prost(string, tag = "1")]
 	pub invoice: ::prost::alloc::string::String,
 }
 /// The response for the `DecodeInvoice` RPC. On failure, a gRPC error status is returned.
+/// `kind` indicates which invoice type was decoded; fields that do not apply to that type
+/// are left empty (e.g. `payment_secret` and `route_hints` are BOLT11-only).
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[cfg_attr(feature = "serde", serde(rename_all = "snake_case"))]
 #[cfg_attr(feature = "serde", serde(default))]
@@ -1173,6 +1175,9 @@ pub struct DecodeInvoiceResponse {
 	/// Whether the invoice has expired.
 	#[prost(bool, tag = "15")]
 	pub is_expired: bool,
+	/// The kind of decoded invoice: "bolt11" or "bolt12".
+	#[prost(string, tag = "16")]
+	pub kind: ::prost::alloc::string::String,
 }
 /// Decode a BOLT12 offer and return its parsed fields.
 /// This does not require a running node — it only parses the offer string.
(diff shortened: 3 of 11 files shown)


benthecarman:
Do we even have a way to get a bolt 12 invoice?

vincenzopalazzo:
> Do we even have a way to get a bolt 12 invoice?

Ah good point, and yes and no! With the new version of ldk-node we can perform a proof of payment (first version of it) with https://github.com/lightningdevkit/ldk-node/pull/733 but people may want to use ldk to decode an bolt12 invoice that they are getting from somewhere else

ldk-server/src/api/decode_invoice.rs
```diff
@@ -20,12 +21,40 @@ use crate::api::decode_features;
...
+
+/// Attempts to decode `invoice` as a hex-encoded BOLT12 invoice.
+///
+/// Unlike offers and BOLT11 invoices, a BOLT12 invoice has no human-readable string
+/// encoding — it is exchanged as raw bytes — so the input is expected to be hex-encoded.
+/// Only the `kind` field is populated for a BOLT12 invoice.
+fn decode_bolt12_invoice(invoice: &str) -> Option<DecodeInvoiceResponse> {
+	let bytes = Vec::<u8>::from_hex(invoice).ok()?;
+	Bolt12Invoice::try_from(bytes).ok()?;
+	Some(DecodeInvoiceResponse { kind: INVOICE_KIND_BOLT12.to_string(), ..Default::default() })
```
benthecarman: whats the point of decoding the bolt12 invoice if we don't fill in any data? We just validate its a bolt12 invoice and thats it?

vincenzopalazzo: Good call, you're right, the first pass only validated the invoice. Fixed in 86d444d: `decode_bolt12_invoice` now fills in the fields that map onto `DecodeInvoiceResponse` -- `destination` (signing pubkey), `payment_hash`, `amount_msat`, `timestamp` (`created_at`), `expiry` (`relative_expiry`), `description`, `fallback_address`, `features`, and `is_expired`. BOLT11-only fields (`payment_secret`, `route_hints`, `currency`, ...) stay empty for BOLT12 invoices.

The `decodes_bolt12_invoice_and_populates_fields` unit test asserts the populated values.

ldk-server/src/api/decode_invoice.rs
```diff
@@ -11,21 +11,68 @@ use std::str::FromStr;
...
+/// left at their default empty values.
+fn decode_bolt12_invoice(invoice: &str) -> Option<DecodeInvoiceResponse> {
+	let bytes = Vec::<u8>::from_hex(invoice).ok()?;
+	let invoice = Bolt12Invoice::try_from(bytes).ok()?;
 
+	let features = decode_features(invoice.invoice_features().le_flags(), |bytes| {
+		Bolt12InvoiceFeatures::from_le_bytes(bytes).to_string()
+	});
+
+	Some(DecodeInvoiceResponse {
```
benthecarman, line 72: can we add to this the blinded paths too?

vincenzopalazzo: **Fixed in ee77107.** `DecodeInvoiceResponse` now carries a `paths` field with the invoice's blinded payment paths (`Bolt12Invoice::payment_paths()`), using the same `BlindedPath` proto type already exposed for decoded offers. The path-to-proto conversion is extracted into a shared `blinded_path_to_proto` helper so offer and invoice decoding stay in sync. `decodes_bolt12_invoice_and_populates_fields` asserts the populated path (hop count, blinding point, introduction node).

ldk-server-grpc/src/api.rs
```diff
@@ -1173,6 +1176,12 @@ pub struct DecodeInvoiceResponse {
 	/// Whether the invoice has expired.
 	#[prost(bool, tag = "15")]
 	pub is_expired: bool,
+	/// The kind of decoded invoice: "bolt11" or "bolt12".
+	#[prost(string, tag = "16")]
+	pub kind: ::prost::alloc::string::String,
```
benthecarman, line 1181: Can we make this a proper enum

e2e-tests/tests/e2e.rs
benthecarman, line 1: can you add a e2e test

e2e-tests/tests/mcp.rs
benthecarman, line 1: can you add a mcp test

benthecarman (changes requested):

tnull (commented):
Hmm,  really not sure if we should make this part of the official API given that AFAIK BOLT12 invoices still don't have a specified encoding format? Yes, hex bytes work for the LDK type, but are we positive this works with other implementations?

I know there has been a lot of previous discussions around that, but should we maybe revisit whether we can add a blessed bech32 encoding for BOLT12 invoices **before** we make it part of our official API? It seems there is a requirement by now, and maybe we should account for that in the spec?

(cc @jkczyz @TheBlueMatt)

vincenzopalazzo:
> Hmm, really not sure if we should make this part of the official API given that AFAIK BOLT12 invoices still don't have a specified encoding format? Yes, hex bytes work for the LDK type, but are we positive this works with other implementations?

Oh, right, I kind of forgot about it that LDK is not going to expose the Bolt12 Invoice, and this is linked to the reason that we do not have a real bech encoding!

Sorry, it completly when off my mind! Probably we should just wait for https://github.com/lightningdevkit/rust-lightning/pull/4297

tnull:
> Oh, right, I kind of forgot about it that LDK is not going to expose the Bolt12 Invoice, and this is linked to the reason that we do not have a real bech encoding!

Yeah, see the discussion over at https://github.com/lightningdevkit/rust-lightning/pull/3800

I think we might want to reconsider that if we find that it's necessary for P-o-P (but I think it just adds `lnp`, right?), but it should probably discussed/reconsidered at the spec level first before we add an API surface for it.

vincenzopalazzo:
The real problem I see here, I think, is not PoP but the migration from a CLN to the ldk-server. Currently, people who are using bolt12 are performing (the few who know what it is) proof of payer or proof that the invoice that they are paying is attached to the offer.

So, I see that there will be the need for this command, but I also see that with the payer proof update to the bolt12, this problem will be gone anyway, without exposing the bolt12 invoice.

However, it will require some time to do this migration

benthecarman:
I do think we will eventually want this. We expose the bolt12 invoice in the events and if you can't decode it, then the value isn't very useful

https://github.com/lightningdevkit/ldk-node/blob/8d6af069c772981b026b8c29a472f3509d68a2bb/src/event.rs#L128

tnull:
> I do think we will eventually want this. We expose the bolt12 invoice in the events and if you can't decode it, then the value isn't very useful
> 
> https://github.com/lightningdevkit/ldk-node/blob/8d6af069c772981b026b8c29a472f3509d68a2bb/src/event.rs#L128

Right, but that means we need a blessed invoice encoding, not just roll our own hex format. Historically @TheBlueMatt has been pushing back on that, but since we already ended up exposing it in our API, we might need to reconsider?

tnull:
So, are we good to close this @vincenzopalazzo, or is there something that we still need to discuss?

PR #216, opened 2026-05-20
Reject unknown config options

benthecarman:
I spent too long debugging why my node wasn't working only to find out I misspelled a config option. Now we'll throw an error if we see a config option we don't recognize.

Fail TOML parsing when unsupported config keys are present so typos do not get silently ignored. Keep the LSPS2 service section typed in all builds, while default builds still parse and ignore it unless the experimental feature is enabled.

### DIFF
--- a/ldk-server-cli/src/main.rs
+++ b/ldk-server-cli/src/main.rs
@@ -563,7 +563,20 @@ async fn main() {
 	}
 
 	let config_path = cli.config.map(PathBuf::from).or_else(get_default_config_path);
-	let config = config_path.as_ref().and_then(|p| load_config(p).ok());
+	let config = match config_path.as_ref() {
+		None => None,
+		Some(path) => {
+			if path.is_file() {
+				let cfg = load_config(path).unwrap_or_else(|e| {
+					eprintln!("Failed to load config file '{}': {}", path.display(), e);
+					std::process::exit(1);
+				});
+				Some(cfg)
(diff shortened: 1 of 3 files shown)


tankyleo:
Nice not sure if it's related, but I also got upset when my ldk-server-client silently dropped the `config.toml` file because I didn't set the `network` field IIRC.

benthecarman:
> Nice not sure if it's related, but I also got upset when my ldk-server-client silently dropped the `config.toml` file because I didn't set the `network` field IIRC.

? can you reproduce that seems weird

lorenzolfm (approved):
ACK 729308c171a14792c9cc961815cecdb26bdb6cee

tankyleo:
> ? can you reproduce that seems weird

Right here, it silently drops the error if the config file failed to  parse.

https://github.com/lightningdevkit/ldk-server/blob/56be35fdead04576d143b39ef548fdd20f794255/ldk-server-cli/src/main.rs#L566

Wasn't sure if worth fixing

benthecarman:
made a fix, good find!

tankyleo (commented):
Let's also reject unknown fields at `ldk-server-client/src/config.rs` ?

benthecarman:
> Let's also reject unknown fields at `ldk-server-client/src/config.rs` ?

done

tankyleo (approved):

PR #218, opened 2026-05-22
Allow unknown config options in client

benthecarman:
Bug accidentally introduced in #216, now tests will catch

We purposely don't parse the entire the config in the client, only the things releveant to the client. When we added the denys in e5eccc1 we now will fail with a normal config because our cli will throw an error for the config options we don't parse in the client.

### DIFF
--- a/e2e-tests/build.rs
+++ b/e2e-tests/build.rs
@@ -54,6 +54,8 @@ fn main() {
 	println!("cargo:rerun-if-changed=../ldk-server/Cargo.toml");
 	println!("cargo:rerun-if-changed=../ldk-server-cli/src");
 	println!("cargo:rerun-if-changed=../ldk-server-cli/Cargo.toml");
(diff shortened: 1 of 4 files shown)


tankyleo:
> Bug accidentally introduced in #216, now tests will catch
> 
> We purposely don't parse the entire the config in the client, only the things releveant to the client. When we added the denys in [e5eccc1](https://github.com/lightningdevkit/ldk-server/commit/e5eccc14029445492a45572cae4578ae113ca72d) we now will fail with a normal config because our cli will throw an error for the config options we don't parse in the client.

Ah my bad, I was still in the old world where we have `ldk-server-config.toml` !

So I thought clients just had a separate configuration file called `config.toml`. I did not understand `config.toml` is now also the config file for `ldk-server` :)

tankyleo (approved):

PR #219, opened 2026-05-22
Include on-chain payments in list payments API

benthecarman:
ListPayments only read from ldk-server's paginated payment store, which is populated from Lightning payment events. On-chain payments are stored by LDK Node directly, so they never appeared in paginated list responses.

Backfill on-chain payments from LDK Node before reading the existing paginated store. This keeps list payments complete until LDK Node owns paginated listing for both Lightning and on-chain payments. Eventually we can remove this when we use the paginated KV store all throughout ldk-node.

Add an e2e test covering an on-chain send appearing in list payments.

### DIFF
--- a/ldk-server/src/api/list_payments.rs
+++ b/ldk-server/src/api/list_payments.rs
@@ -10,6 +10,7 @@
 use std::sync::Arc;
 
 use bytes::Bytes;
+use ldk_node::payment::{PaymentDetails, PaymentKind};
 use ldk_server_grpc::api::{ListPaymentsRequest, ListPaymentsResponse};
 use ldk_server_grpc::types::{PageToken, Payment};
 use prost::Message;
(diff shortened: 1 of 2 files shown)


ldk-server/src/api/list_payments.rs
```diff
@@ -64,3 +71,32 @@ pub(crate) async fn handle_list_payments_request(
 	};
 	Ok(response)
 }
+
+// TODO: Delete this temporary bridge when on-chain and Lightning payments are served from the
+// same paginated LDK Node source.
+fn sync_onchain_payments_to_paginated_store(context: &Context) -> Result<(), LdkServerError> {
+	for payment_details in context.node.list_payments().into_iter().filter(is_onchain_payment) {
```
lorenzolfm: theres a list_payments_with_filter api already available btw


```suggestion
    for payment_details in context.node.list_payments_with_filter(|p| matches!(p.kind, PaymentKind::Onchain { .. }))
```

tnull (commented):
Hmm, I'm confused why we need this, isn't the plan to enable pagination on LDK Node's payment store and to drop the separate paginated store in Server entirely?

benthecarman:
> Hmm, I'm confused why we need this, isn't the plan to enable pagination on LDK Node's payment store and to drop the separate paginated store in Server entirely?

Yeah this is a temporary fix 

> Eventually we can remove this when we use the paginated KV store all throughout ldk-node.

ldk-server/src/api/list_payments.rs
benthecarman: fixed, thanks

benthecarman:
just gonna leave the bug until fixed it in ldk-node (will be fixed before we release)

PR #220, opened 2026-05-23
Bound gRPC frame sizes

benthecarman:
Reject oversized server-streaming gRPC messages as soon as their frame
header is available. Also reject compressed stream frames consistently
and avoid unchecked frame length arithmetic before slicing buffers.

Read unary client responses incrementally and reject bodies above the
configured limit before protobuf decoding. Preallocate only after a
valid Content-Length has been checked.

tested with all the graph clis and they still fit within the limit

### DIFF
--- a/ldk-server-client/src/client.rs
+++ b/ldk-server-client/src/client.rs
@@ -436,6 +436,7 @@ impl LdkServerClient {
 		&self, request: &Rq, method: &str,
 	) -> Result<Rs, LdkServerError> {
 		let grpc_body = encode_grpc_frame(&request.encode_to_vec()).to_vec();
+		let content_length = grpc_body.len().to_string();
 
 		let url = format!("https://{}{}{}", self.base_url, GRPC_SERVICE_PREFIX, method);
 		let auth_header = self.compute_auth_header(&grpc_body);
@@ -444,6 +445,7 @@ impl LdkServerClient {
 			.client
 			.post(&url)
 			.header("content-type", "application/grpc+proto")
+			.header("content-length", content_length)
 			.header("te", "trailers")
 			.header("x-auth", auth_header)
 			.body(grpc_body)
@@ -476,6 +478,7 @@ impl LdkServerClient {
 		&self, request: &Rq, method: &str,
 	) -> Result<GrpcStream<Rs>, LdkServerError> {
 		let grpc_body = encode_grpc_frame(&request.encode_to_vec()).to_vec();
+		let content_length = grpc_body.len().to_string();
 
 		let url = format!("https://{}{}{}", self.base_url, GRPC_SERVICE_PREFIX, method);
 		let auth_header = self.compute_auth_header(&grpc_body);
@@ -486,6 +489,7 @@ impl LdkServerClient {
 				HyperRequest::post(&url)
 					.version(Version::HTTP_2)
 					.header("content-type", "application/grpc+proto")
+					.header("content-length", content_length)
 					.header("te", "trailers")
 					.header("x-auth", auth_header)
 					.body(HyperBody::from(grpc_body))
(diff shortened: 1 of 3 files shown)


ldk-server-client/src/client.rs
```diff
@@ -71,6 +71,16 @@ use crate::error::LdkServerErrorCode::{
 
 type StreamingClient = HyperClient<HttpsConnector<hyper::client::HttpConnector>, HyperBody>;
 
+const GRPC_FRAME_HEADER_LEN: usize = 5;
+
+// Applies to complete unary gRPC responses. The server applies the same cap to unary request
+// bodies before protobuf decoding.
+const MAX_GRPC_UNARY_RESPONSE_LEN: usize = 10 * 1024 * 1024;
```
lorenzolfm, line 78: wdyt about having the unit explicit in the const name? `_BYTES` instead of `_LEN`?

ldk-server-client/src/client.rs
```diff
@@ -508,6 +515,42 @@ impl LdkServerClient {
...
+		let len = payload.len().checked_add(chunk.len()).ok_or_else(|| {
+			LdkServerError::new(InternalError, "gRPC unary response body length overflow")
+		})?;
+		check_grpc_unary_response_len(len as u64)?;
+		payload.extend_from_slice(&chunk);
+	}
+	Ok(payload)
+}
+
+fn check_grpc_unary_response_len(len: u64) -> Result<(), LdkServerError> {
```
lorenzolfm, line 545: if you drop the content_size path, you can change this to receive `usize` and avoid the `as` castings both on L535 and L542

ldk-server-client/src/client.rs
```diff
@@ -508,6 +515,42 @@ impl LdkServerClient {
...
 
+async fn read_grpc_unary_response_body(
+	mut response: reqwest::Response,
+) -> Result<Vec<u8>, LdkServerError> {
+	let capacity = if let Some(content_length) = response.content_length() {
+		check_grpc_unary_response_len(content_length)?;
+		content_length as usize
+	} else {
+		0
+	};
```
lorenzolfm, line 530: I think the `Some` path is essentially useless as ldk-server doesn't set the content length header. It actually could't even if it wanted to. In gRPC the body streams and you can't compute a length.

Also, maybe worth noting that having `Vec::with_capacity(0)` hardcoded it's kinda bad because it will always have to allocate. Given the context where this is running it seems like not that big of a deal. You could initialize for something different than zero, or maybe use some BytesMut trickery to alloc exactly what you need.

joostjager:
We hand-rolled gRPC to reduce exposure to supply chain risk, but I think this PR demonstrates that it is a trade-off because we might introduce vulnerabilities ourselves. Hopefully we can keep the gRPC feature set small enough that the balance does not shift unfavourably.

ldk-server-client/src/client.rs
benthecarman: `Vec::with_capacity(0)` is the same as `Vec::new()` so this is still an improvement from the pervious version but yeah, I should make sure we return the content length from the server

ldk-server-client/src/client.rs
benthecarman: i think that is implied

ldk-server-client/src/client.rs
benthecarman: if we do that then we need to either cast the `content_length` before validating it which is unsafe, or do a `try_into` which ends up being more ugly than these casts.

benthecarman:
added 2 commits where we make sure we always send the content length on both client and server to make the handling cleaner

valentinewallace (commented):
I'm not sure I'm the right person to review this, don't have a lot of context on gRPC. Claude seems to think this is fine.

tnull (approved):

PR #221, opened 2026-05-23
Sanitize error messages for terminal

benthecarman:
We call `sanitize_for_terminal` on the output for successful responses but not on error outputs. We now call it on errors to protect about potentially malicous error messages that could try to manipulate the terminal output.


lorenzolfm (approved):
ACK 4eabaf504fab689c612f9b07df68b9b8fd50290c

valentinewallace (approved):

PR #227, opened 2026-06-01
Add Codex agent instructions symlink

benthecarman:
Create AGENTS.md as a symlink to CLAUDE.md so Codex can load the repository guidance from the existing agent instructions file.


tnull (approved):
Very annoying that claude won't read it automatically.

PR #228, opened 2026-06-08
Display compressed public keys in info logs

tankyleo:
```
    Display compressed public keys in info logs

    Uncompressed public keys are not very helpful, especially since our
    `list-peers` command displays compressed public keys for node ids.
```


benthecarman (approved):
Good find!

benthecarman:
ci not working but ran everything locally and passed

PR #229, opened 2026-06-08
Add crate version and commit hash to ldk-server

benthecarman:
We will now log the version number and commit hash when starting up. Also will give the full commit hash when doing ldk-server --version to better help debugging things.

Also does the same to the cli

### DIFF
--- a/ldk-server/build.rs
+++ b/ldk-server/build.rs
@@ -1,0 +1,58 @@
+// This file is Copyright its original authors, visible in version control
+// history.
+//
+// This file is licensed under the Apache License, Version 2.0 <LICENSE-APACHE
+// or http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your option.
+// You may not use this file except in accordance with one or both of these
+// licenses.
+
+use std::env;
+use std::path::Path;
+use std::process::Command;
+
+fn main() {
+	println!("cargo:rerun-if-changed=build.rs");
+	println!("cargo:rerun-if-env-changed=GIT_HASH");
+
+	// Re-run the build script whenever the checked-out commit changes so the
+	// embedded hash never goes stale. HEAD lives in the (possibly worktree-local)
+	// git dir, while branch refs and packed-refs live in the common git dir.
+	if let Some(git_dir) = git_output(&["rev-parse", "--git-dir"]) {
+		watch_path(&Path::new(&git_dir).join("HEAD"));
+	}
+	if let Some(common_dir) = git_output(&["rev-parse", "--git-common-dir"]) {
+		let common_dir = Path::new(&common_dir);
+		watch_path(&common_dir.join("packed-refs"));
+		// If HEAD points at a ref, watch that ref file too (it changes on commit).
+		// Watch it even when it does not exist yet: packed refs become loose
+		// files on the next commit, and Cargo can detect that creation.
+		if let Some(ref_path) = git_output(&["symbolic-ref", "-q", "HEAD"]) {
+			watch_path(&common_dir.join(ref_path));
+		}
+	}
+
+	let git_hash = git_output(&["rev-parse", "HEAD"])
+		.or_else(|| env::var("GIT_HASH").ok())
+		.unwrap_or_else(|| "unknown".to_string());
+	println!("cargo:rustc-env=GIT_HASH={git_hash}");
+}
+
+/// Runs `git` with the given args, returning the trimmed stdout on success or
+/// `None` if git is unavailable, exits non-zero, or produces no output.
+fn git_output(args: &[&str]) -> Option<String> {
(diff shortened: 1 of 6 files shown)


ldk-server/src/main.rs
```diff
@@ -241,7 +242,7 @@ fn main() {
 	let (event_sender, _) = broadcast::channel::<EventEnvelope>(1024);
 	let (shutdown_tx, shutdown_rx) = tokio::sync::watch::channel(false);
 
-	info!("Starting up...");
+	info!("Starting ldk-server version {FULL_VERSION}");
```
tankyleo, line 244: Should we also do this when we rotate the logs ?

ldk-server/build.rs
```diff
@@ -0,0 +1,26 @@
...
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your option.
+// You may not use this file except in accordance with one or both of these
+// licenses.
+
+use std::env;
+use std::process::Command;
+
+fn main() {
+	println!("cargo:rerun-if-changed=build.rs");
+	println!("cargo:rerun-if-env-changed=GIT_HASH");
```
tankyleo, line 16: Can we make sure to update `GIT_HASH` if `HEAD` changes after the first from-clean build. Here it seems `HEAD` could change, and we'd ship a binary with a wrong rev ?

Or perhaps `cargo install` always re-runs `build.rs` so we might be good here.

tankyleo: Also trying to understand how this triggers. After `build.rs` sets `GIT_HASH`, what other thing could cause `GIT_HASH` to change, and trigger `build.rs` again ?

ldk-server/src/main.rs
benthecarman: can do after #189

ldk-server/build.rs
benthecarman: Made it so it watches the git hash file so if it changes then we recalc

tankyleo (commented):
vibed with codex, pushed the suggestions in a commit, let me know what you think @benthecarman 

The findings:

```
_ Findings

  1. High: inherited GIT_HASH overrides the real repository hash in both build scripts: ldk-server/build.rs:33, ldk-server-cli/build.rs:33. Your report is correct. If the shell exports GIT_HASH, the script skips cargo:rustc-env, so env!("GIT_HASH") compiles the caller_s stale value. I reproduced:
     GIT_HASH=stale-from-env cargo run -q -p ldk-server-cli -- --version -> ldk-server-cli 0.1.0 (stale-from-env), and same for ldk-server.

     Suggested fix: always compute from git rev-parse HEAD when git is available, then emit cargo:rustc-env=GIT_HASH=.... Only fall back to the environment when git is unavailable, or use a package-specific override variable.

  2. Medium: packed branch refs can also leave the hash stale: ldk-server/build.rs:28, ldk-server-cli/build.rs:28, ldk-server/build.rs:54. If the current branch exists only in .git/packed-refs, watch_if_exists skips .git/refs/heads/<branch>. The next commit creates that loose ref, but HEAD and
  packed-refs do not change, so Cargo does not rerun the build script. I reproduced this in a throwaway clone: after an empty commit advanced HEAD from 5d6e906... to 80648b9..., ldk-server-cli --version still printed 5d6e906....

     Suggested fix: print cargo:rerun-if-changed for the symbolic ref path even when it does not exist yet. I validated that Cargo then reruns when Git creates the file.

  Checks run: cargo fmt --all -- --check, cargo check -p ldk-server, cargo check -p ldk-server-cli, configured cargo clippy --all-features -- -D warnings -A clippy::drop_non_drop, cargo test -p ldk-server-cli, and cargo test -p ldk-server all passed.
```

benthecarman:
@tankyleo thanks I like those changes. Squashed them into my commits and added you as co-author

tankyleo (commented):
What do we want to do with docker? Had codex produce this commit below, could also read the external GIT_HASH env var
```diff
Date:   Wed Jun 24 16:38:56 2026 +0000

    Let Docker builds read Git metadata

    Stop excluding .git from the Docker build context. The version build
    scripts can then discover HEAD during image builds, matching local Cargo
    builds when the source comes from a normal checkout.

    Keep the existing GIT_HASH environment fallback for builds that do not
    have Git metadata available. That fallback remains the last resort after
    git rev-parse HEAD fails, and Cargo still reruns the build scripts when
    the fallback value changes.

    AI-Generated-By: OpenAI Codex

diff --git a/.dockerignore b/.dockerignore
index a4ec515..d21fbcc 100644
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,6 +1,5 @@
 target/
 **/target/
-.git/
 .gitignore
 .github/
 .claude/
```

benthecarman:
I'd rather not include the whole git dir in the build. Made it so we can pass the env var as an arg. This way we can also do like `0.3.0` instead of a commit hash so the build looks more like an official release

tankyleo (approved):

PR #230, opened 2026-06-09
Bump ldk-node dependency to `00dba456a10cb60172eaa74e0a8b11c5e0473dcc`

enigbe:
## What this PR does:

Bumps the `ldk-node` dependency to revision [47dad6d](https://github.com/lightningdevkit/ldk-node/commit/47dad6d909af0fbb53e76d07740c04df5abdde3b), which, among other changes, contains the prerequisite for exposing `NodeFeatures` to `ldk-server` callers via the node `status` endpoint. We need this for callers interested in querying the features supported by the node. 

Notable upstream changes adapted here
- `Bolt11Jit` merged into Bolt11: ldk-node folded the JIT payment variant into `Bolt11`, adding a `counterparty_skimmed_fee_msat` field. The `Bolt11Jit` proto message, its serde attribute in `build.rs`, and the corresponding proto_adapter arm are all removed. The original proto tag numbers for Bolt12Offer (4), Bolt12Refund (5), and Spontaneous (6) are preserved — tag 3 is reserved to avoid wire-compatibility issues with previously-serialized `Bolt11Jit` data.

- `PaymentForwarded` event restructured around `HTLCLocator`: The six flat channel/node/user-channel-id fields are replaced by `prev_htlcs` and `next_htlcs` (`Vec<HTLCLocator>`), supporting multi-HTLC forwards. The legacy scalar fields are retained in the proto as deprecated, populated from the first element of each vec for backward compatibility (not sure if this is the correct way to handle this).

### DIFF
--- a/ldk-server-grpc/src/proto/types.proto
+++ b/ldk-server-grpc/src/proto/types.proto
@@ -33,10 +33,9 @@ message PaymentKind {
   oneof kind {
     Onchain onchain = 1;
     Bolt11 bolt11 = 2;
-    Bolt11Jit bolt11_jit = 3;
-    Bolt12Offer bolt12_offer = 4;
-    Bolt12Refund bolt12_refund = 5;
-    Spontaneous spontaneous = 6;
+    Bolt12Offer bolt12_offer = 3;
+    Bolt12Refund bolt12_refund = 4;
+    Spontaneous spontaneous = 5;
   }
 }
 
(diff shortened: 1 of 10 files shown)


benthecarman:
Can you bump to `00dba456a10cb60172eaa74e0a8b11c5e0473dcc`? This will have some of the async persistence fixes and includes the new kv stores that I want to add

ldk-server-grpc/src/proto/types.proto
```diff
@@ -30,10 +30,12 @@ message Payment {
 }
 
 message PaymentKind {
+  reserved 3;
+  reserved "bolt11_jit";
```
benthecarman: we aren't worrying about backwards compat yet so you can remove this

ldk-server-grpc/src/proto/types.proto
```diff
@@ -219,27 +208,58 @@ enum Network {
...
+//
+// A forwarded payment can involve multiple incoming and outgoing HTLCs, e.g. when acting as a
+// trampoline router. The `prev_htlcs` and `next_htlcs` fields are the canonical representation of
+// the HTLCs associated with this forwarding event. Their indices do not imply pairwise
+// correspondence.
+//
 // See more: https://docs.rs/ldk-node/latest/ldk_node/enum.Event.html#variant.PaymentForwarded
 message ForwardedPayment{
   // The channel id of the incoming channel between the previous node and us.
+  // Deprecated: use `prev_htlcs` instead. This is populated from `prev_htlcs[0]`
```
benthecarman: if these fields are deprecated it's probably best if we drop them all

ldk-server-grpc/src/proto/types.proto
```diff
@@ -81,32 +83,19 @@ message Bolt11 {
...
-  // See [`LdkChannelConfig::accept_underpaying_htlcs`](https://docs.rs/lightning/latest/lightning/util/config/struct.ChannelConfig.html#structfield.accept_underpaying_htlcs)
-  // for more information.
-  LSPFeeLimits lsp_fee_limits = 4;
 
   // The value, in thousands of a satoshi, that was deducted from this payment as an extra
   // fee taken by our channel counterparty.
   //
-  // Will only be `Some` once we received the payment.
-  optional uint64 counterparty_skimmed_fee_msat = 5;
+  // Replaces the former `PaymentKind.bolt11_jit.counterparty_skimmed_fee_msat`
```
benthecarman: we dont need to mention the old format, we haven't released yet so no one knows the former one

ldk-server-grpc/src/proto/types.proto
enigbe: Noted. I'll remove all code that considered backwards compatibility.

enigbe:
> Can you bump to `00dba456a10cb60172eaa74e0a8b11c5e0473dcc`? This will have some of the async persistence fixes and includes the new kv stores that I want to add

Sure thing. Working on this now.

enigbe:
Hi @benthecarman 
This is ready for another look.

ldk-server-grpc/src/proto/types.proto
benthecarman: this doc still talks about old format

benthecarman (commented):
otherwise lgtm

enigbe:
> otherwise lgtm

Cleaned up the documentation still referencing `Bolt11Jit`. Should be good to go now.
Thanks for the review @benthecarman

enigbe:
Added a couple more commits to address failing tests. If you consider them out of scope for this PR, I am happy to move them to another.

benthecarman (approved):
thanks!

PR #232, opened 2026-06-14
feat: add event kind filtering to event subscriptions

### LINKED ISSUE #225: Subscribe to specfic event kinds
We currently only allow for subscribing to all events. We should also allow to subscribing to smaller subsets like just payments events or just channel events, etc

JemimahEkong:
## Summary

This PR adds support for subscribing to specific event kinds instead of always receiving all events.

Previously, all clients received every event emitted by the server. This change introduces optional filtering so clients can subscribe to only relevant event categories.

### Proto changes

* Added `EventKind` enum:

  * UNSPECIFIED = 0
  * PAYMENT = 1
  * CHANNEL = 2
* Added `repeated EventKind event_kinds` field to `SubscribeEventsRequest`

### Server changes

* Added `event_kind()` helper to map `EventEnvelope → Option<EventKind>`
* Added per-subscriber filtering logic in event streaming handler
* `EventKind::Unspecified` is stripped and treated as no filter (equivalent to all events)
* Empty filter continues to return all events (backward compatible behavior)

### Client changes

* Restored `subscribe_events()` as backward-compatible API (no arguments)
* Added `subscribe_events_filtered(&[EventKind])` for filtered subscriptions
* Re-exported `EventKind` for external usage

### Tests

* Added unit tests for:

  * empty filter (all events)
  * payment-only filtering
  * channel-only filtering
  * multiple event kinds
  * EventKind::Unspecified handling
* Updated all existing e2e tests to maintain backward compatibility

## Backward Compatibility

* Existing clients using `subscribe_events()` are unaffected
* Empty filter continues to return all events
* Wire format is unchanged

## Notes

* `api.rs` includes a hand-synced update to mirror proto definitions
* Proper regeneration should be done via protobuf build tooling when available
* Event filtering is performed per subscriber without modifying broadcast behavior

## Status

Ready for review. CI will validate compilation and test correctness.
Closes #225

### DIFF
--- a/ldk-server-grpc/src/api.rs
+++ b/ldk-server-grpc/src/api.rs
@@ -1233,10 +1233,55 @@ pub struct DecodeOfferResponse {
 	#[prost(bool, tag = "12")]
 	pub is_expired: bool,
 }
+// NOTE: This section (EventKind + SubscribeEventsRequest) is hand-synced with
+// ldk-server-grpc/src/proto/api.proto. If you modify the proto definition, re-run
+//     RUSTFLAGS="--cfg genproto" cargo build -p ldk-server-grpc
+// to regenerate this file. Manual edits will be overwritten.
+
+/// EventKind represents the category of events a client can subscribe to.
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serde", serde(rename_all = "snake_case"))]
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, PartialOrd, Ord, ::prost::Enumeration)]
+#[repr(i32)]
+pub enum EventKind {
+	/// Default value, not used for filtering.
+	Unspecified = 0,
+	/// Payment events (received, successful, failed, forwarded, claimable).
+	Payment = 1,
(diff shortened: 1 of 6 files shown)


ldk-server-grpc/src/api.rs
```diff
@@ -1233,10 +1233,55 @@ pub struct DecodeOfferResponse {
 	#[prost(bool, tag = "12")]
 	pub is_expired: bool,
 }
+// NOTE: This section (EventKind + SubscribeEventsRequest) is hand-synced with
+// ldk-server-grpc/src/proto/api.proto. If you modify the proto definition, re-run
+//     RUSTFLAGS="--cfg genproto" cargo build -p ldk-server-grpc
+// to regenerate this file. Manual edits will be overwritten.
```
benthecarman, line 1239: why are you doing this? This file is supposed to be generated

benthecarman (changes requested):
Rather than everything being loaded into `subscribe_events` I'd prefer if we left that one alone and instead added a `subscribe_channel_events` and `subscribe_payment_events`.

JemimahEkong:
Noted. Will work on it.

benthecarman:
@JemimahEkong any progress here?

PR #233, opened 2026-06-15
feat: allow setting custom preimage in spontaneous-send

### LINKED ISSUE #226: Allow setting custom preimage in `spontaneous-send`
ldk-node supports setting a custom preimage instead of letting the node generate it. We should expose this in the api

f3r10:
address #226 

## Summary
- Adds an optional `preimage` field to `SpontaneousSendRequest`, allowing callers to supply a custom 32-byte hex preimage instead of having the node generate one randomly
- Dispatches to the appropriate ldk-node method (`send_with_preimage` or `send_with_preimage_and_custom_tlvs`) based on whether preimage and/or custom TLVs are present
- Exposes the flag in the CLI (`--preimage`) and MCP schema
- Adds an e2e test verifying the receiver observes a payment hash equal to
    SHA256 of the provided preimage

### DIFF
--- a/e2e-tests/tests/e2e.rs
+++ b/e2e-tests/tests/e2e.rs
@@ -32,6 +32,7 @@ use ldk_server_client::ldk_server_grpc::events::{
 use ldk_server_client::ldk_server_grpc::types::{
 	bolt11_invoice_description, Bolt11InvoiceDescription,
 };
+use ldk_server_grpc::types::payment_kind;
 
 const EVENT_TIMEOUT: Duration = Duration::from_secs(15);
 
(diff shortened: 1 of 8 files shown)


e2e-tests/tests/e2e.rs
```diff
@@ -32,7 +32,7 @@ use ldk_server_client::ldk_server_grpc::events::{
 use ldk_server_client::ldk_server_grpc::types::{
 	bolt11_invoice_description, Bolt11InvoiceDescription,
 };
-
+use ldk_server_grpc::types::payment_kind;
```
benthecarman, line 35: nit: can we have a new line between the imports and const

ldk-server-cli/src/main.rs
```diff
@@ -320,6 +320,8 @@ enum Commands {
 			help = "Custom TLV record to attach, format: <type_num>:<hex_value>. Repeatable. type_num must be >= 65536."
 		)]
 		custom_tlvs: Vec<(u64, Vec<u8>)>,
+		#[arg(long)]
```
benthecarman: please add help text

benthecarman (changes requested):

ldk-server/src/api/spontaneous_send.rs
```diff
@@ -27,19 +29,48 @@ pub(crate) async fn handle_spontaneous_send_request(
...
+			<[u8; 32]>::from_hex(&p).map(PaymentPreimage).map_err(|_| {
+				LdkServerError::new(
+					InvalidRequestError,
+					"Invalid preimage, must be a 32-byte hex string.".to_string(),
+				)
+			})
+		})
+		.transpose()?;
+
+	let custom_tlvs: Vec<_> = request.custom_tlvs.iter().map(proto_to_node_custom_tlv).collect();
```
Anyitechs: No strong opinion here, especially as I believe there's really no runtime cost in iterating and mapping over an empty vector. But since we're creating the vector and immediately checking if it's empty in the `match` statement below. 

I'm wondering if we could do something like this instead?

```
let custom_tlvs: Option<Vec<CustomTlvRecord>> = if request.custom_tlvs.is_empty() {
  None
} else {
  Some(request.custom_tlvs.iter().map(proto_to_node_custom_tlv).collect())
};

let payment_id = match (preimage, custom_tlvs) {
		(None, None) => context.node.spontaneous_payment().send(
			request.amount_msat,
			node_id,
			route_parameters,
		)?,
		(None, Some(custom_tlvs)) => context.node.spontaneous_payment().send_with_custom_tlvs(
			request.amount_msat,
			node_id,
			route_parameters,
			custom_tlvs,
		)?,
		(Some(preimage), None) => context.node.spontaneous_payment().send_with_preimage(
			request.amount_msat,
			node_id,
			preimage,
			route_parameters,
		)?,
		(Some(preimage), Some(custom_tlvs)) => context
			.node
			.spontaneous_payment()
			.send_with_preimage_and_custom_tlvs(
				request.amount_msat, 
                 node_id, 
                 custom_tlvs, 
                 preimage, 
                 route_parameters,
			)?,
};

```

e2e-tests/tests/e2e.rs
Anyitechs: @f3r10 Looks like you missed this in your latest push.

Anyitechs (commented):
Thanks! 

The CI is currently failing because the proto files are now out of date and needs to be updated.

ldk-server/src/api/spontaneous_send.rs
benthecarman: yeah this makes it a bit more readable

benthecarman:
@f3r10 any progress here?

ldk-server/src/api/spontaneous_send.rs
f3r10: thanks @Anyitechs 👍

ldk-server-cli/src/main.rs
```diff
@@ -819,6 +822,7 @@ async fn main() {
 			max_path_count,
 			max_channel_saturation_power_of_half,
 			custom_tlvs,
+			preimage,
 		} => {
 			let amount_msat = amount.to_msat();
```
benthecarman, line 828: can we do some validation on the preimage before, we should check that it is 64 chars of hex

benthecarman (changes requested):
looks good, minus one last thing

ldk-server-cli/src/main.rs
f3r10: I added a custom preimage type to check that part with clap @benthecarman

benthecarman (approved):

PR #234, opened 2026-06-17
Test all features in CI

benthecarman:
We weren't running our tests with all features on CI. This left one test to be left out and did not pass. Fixes the test by removing the whole LSPS2 config section instead of just the header and enables --all-features for our CI tests.

### DIFF
--- a/ldk-server/src/util/config.rs
+++ b/ldk-server/src/util/config.rs
@@ -1347,8 +1347,22 @@ mod tests {
 
 		#[cfg(feature = "experimental-lsps2-support")]
 		{
-			validate_missing!(
-				"[liquidity.lsps2_service]",
+			let toml_config = r#"
+				[node]
+				network = "regtest"
+
+				[bitcoind]
+				rpc_address = "127.0.0.1:8332"
+				rpc_user = "bitcoind-testuser"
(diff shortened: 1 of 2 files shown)


ldk-server/src/util/config.rs
```diff
@@ -1347,8 +1347,16 @@ mod tests {
...
+			let mut toml_config = DEFAULT_CONFIG.to_string();
+			// Remove the full table so the file remains valid TOML while omitting the service config.
+			toml_config = remove_config_section(&toml_config, "[liquidity.lsps2_service]");
+			fs::write(storage_path.join(config_file_name), &toml_config).unwrap();
+			let result = load_config(&args_config);
+			assert!(result.is_err());
+			let err = result.unwrap_err();
+			assert_eq!(err.kind(), io::ErrorKind::InvalidInput);
+			assert_eq!(
+				err.to_string(),
```
jkczyz, line 1365: Rather than using `DEFAULT_CONFIG`, could we simply inline a minimal config that doesn't have a `[liquidity.lsps2_service]` section? That way we don't need to define `remove_config_section`.

benthecarman: fixed

jkczyz (approved):

PR #235, opened 2026-06-18
grpc: add claim_deadline to PaymentClaimable event

JssDWt:
add claim_deadline to PaymentClaimable event, so that it's easier to determine how much time we have for security in some out-of-lightning protocol using hodl invoices.


benthecarman (approved):
looks good, thanks!

PR #239, opened 2026-07-02
Implement MCP ping method

benthecarman:
Had fable scan the codebase and call out anything we should add. It refused and fell back to opus but at least it found this.

The MCP spec requires receivers to answer a ping request with an empty result object. Previously every method other than initialize, tools/list, and tools/call fell through to a METHOD_NOT_FOUND error, so a client that health-checks with ping would conclude the server was dead and drop the connection.

This change was made with the assistance of Claude Code.

### DIFF
--- a/ldk-server-mcp/CLAUDE.md
+++ b/ldk-server-mcp/CLAUDE.md
@@ -35,7 +35,7 @@ src/
 - **Version**: `2025-11-25`
 - **Spec**: https://spec.modelcontextprotocol.io/
 - **Transport**: stdio (one JSON-RPC 2.0 message per line)
-- **Methods implemented**: `initialize`, `tools/list`, `tools/call`
+- **Methods implemented**: `initialize`, `tools/list`, `tools/call`, `ping`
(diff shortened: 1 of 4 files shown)


ldk-server-mcp/CLAUDE.md
```diff
@@ -35,7 +35,7 @@ src/
 - **Version**: `2025-11-25`
 - **Spec**: https://spec.modelcontextprotocol.io/
 - **Transport**: stdio (one JSON-RPC 2.0 message per line)
-- **Methods implemented**: `initialize`, `tools/list`, `tools/call`
+- **Methods implemented**: `initialize`, `tools/list`, `tools/call`, `ping`
```
tnull, line 38: Codex:

> P3 - /home/tnull/workspace/ldk-server-pr-239/ldk-server-mcp/README.md:104 still lists only initialize, tools/list, and tools/call. Since PR 239 adds ping and updates CLAUDE.md, the public crate README should include ping too.

tnull (commented):
One nit, otherwise LGTM

tnull (approved):

PR #240, opened 2026-07-03
Optimize paginated payment listing queries

joostjager:
This replaces the SQLite pagination index with one that matches the actual paginated store access pattern: namespace filters first, followed by `creation_time DESC, key ASC` for cursor ordering. The previous `creation_time`-only index did not fit these queries well and could scan unrelated namespaces or require extra sorting.

It also adds a required `list_values` paginated store method and implements it in `SqliteStore` with a single query returning keys and serialized values. `ListPayments` and `ListForwardedPayments` now decode those returned values directly, avoiding the previous page-level N+1 pattern where listing 100 items could require 101 SQLite queries.

### DIFF
--- a/ldk-server/src/api/list_forwarded_payments.rs
+++ b/ldk-server/src/api/list_forwarded_payments.rs
@@ -28,7 +28,7 @@ pub(crate) async fn handle_list_forwarded_payments_request(
 	let page_token = request.page_token.map(|p| (p.token, p.index));
 	let list_response = context
 		.paginated_kv_store
-		.list(
+		.list_values(
 			FORWARDED_PAYMENTS_PERSISTENCE_PRIMARY_NAMESPACE,
 			FORWARDED_PAYMENTS_PERSISTENCE_SECONDARY_NAMESPACE,
 			page_token,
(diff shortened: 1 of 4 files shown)


tnull (commented):
We're about to drop the SQLite / local forwarded payment store from Server as we're close to finally add all parts to LDK Node. So I don't think it makes sense to make changes to it right now?

joostjager:
Opened this PR because I still had the code lying around from an older analysis, but I wasn’t fully up to date with what had happened in `ldk-node`.

The main thing to distill from it is probably this question on `ldk-node#959`, if it hasn’t already been answered elsewhere: https://github.com/lightningdevkit/ldk-node/pull/959#discussion_r3518607491

One meta note: the split between `ldk-node` and `ldk-server` does make this kind of thing a bit harder to track from the `ldk-server` side.

tnull:
> One meta note: the split between `ldk-node` and `ldk-server` does make this kind of thing a bit harder to track from the `ldk-server` side.

Yes, well, that split is exactly what we're trying to reduce currently, i.e., move any remaining custom 'logic' into Node so that Server is really just responsible for the API, dealing with the daemon etc.

joostjager:
Once `ldk-server` is cleaned up again and this stateful logic has moved into `ldk-node`, I think it would be good to define the boundary strictly so we avoid future work and rework like this.

And I still think keeping this split across two repos introduces unnecessary visibility and change friction, but you all know my stance on that.

PR #243, opened 2026-07-07
Expose Postgres storage backend

benthecarman:
Allow ldk-server to configure LDK Node wallet state, channel state, and payment history through PostgreSQL without a feature-gated server build. Keep local disk storage for server-owned files and forwarded-payment history, and document the backup and per-network isolation boundaries.

Refuse switching between SQLite and PostgreSQL after initialization. The local marker records only the backend type; it does not detect a changed PostgreSQL database or table, or missing state.

Limit PostgreSQL certificate reads to 1 MiB and test oversized and unbounded inputs. Remove the Getting Started claim that no other external dependencies are required.

Validation: formatting, default and all-feature tests, and Clippy passed.

AI-assisted-by: OpenAI Codex

### DIFF
--- a/contrib/ldk-server-config.toml
+++ b/contrib/ldk-server-config.toml
@@ -28,7 +28,14 @@ alias = "ldk_server"                          # Lightning node alias
 
 # Storage settings
 [storage.disk]
-dir_path = "/tmp/ldk-server/"                 # Path for LDK and BDK data persistence, optional, defaults to ~/Library/Application Support/ldk-server/ on macOS, ~/.ldk-server/ on Linux
+dir_path = "/tmp/ldk-server/"                 # Path for local ldk-server data, optional, defaults to ~/Library/Application Support/ldk-server/ on macOS, ~/.ldk-server/ on Linux
+
+# Optional LDK Node PostgreSQL storage for wallet and channel state.
+#[storage.postgres]
+#connection_string = "postgresql://postgres:postgres@localhost:5432" # PostgreSQL connection string. Do not include dbname if db_name is set.
+#db_name = "ldk_db"                         # Optional database name.
+#kv_table_name = "ldk_data"                 # Optional KV table name.
+#certificate_path = "/path/to/postgres-ca.pem" # Optional CA certificate PEM file for TLS PostgreSQL connections.
 
 [log]
 level = "Debug"                               # Log level (Error, Warn, Info, Debug, Trace)
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -100,8 +100,32 @@ cooldown_secs = 3600
 
 ### `[storage.disk]`
 
-Where persistent data is stored. Defaults to `~/.ldk-server/` on Linux and
-`~/Library/Application Support/ldk-server/` on macOS.
+Where local ldk-server data is stored. Defaults to `~/.ldk-server/` on Linux and
+`~/Library/Application Support/ldk-server/` on macOS. This directory is still used for
+the node mnemonic, API key, TLS material, logs, and ldk-server's payment history when
+LDK Node wallet/channel state uses PostgreSQL.
+
+### `[storage.postgres]`
+
+Optional PostgreSQL storage for LDK Node wallet and channel state.
+
+```toml
+[storage.postgres]
+connection_string = "postgresql://postgres:postgres@localhost:5432"
+db_name = "ldk_db"
+kv_table_name = "ldk_data"
+certificate_path = "/path/to/postgres-ca.pem"
+```
+
+Only `connection_string` is required. `db_name`, `kv_table_name`, and `certificate_path`
+are optional. If `db_name` is set, do not also include a database name in the connection
+string. If `certificate_path` is set, the file must contain a PEM-encoded CA certificate
+for TLS PostgreSQL connections.
+
+Storage migration is not supported. ldk-server refuses to start with PostgreSQL when an existing
+`ldk_node_data.sqlite` file is present. After the first successful PostgreSQL node build,
+ldk-server creates `<network_dir>/ldk_node_postgres.lock` and refuses to start with SQLite while
+that file exists.
 
 ### `[log]`
 
@@ -204,6 +228,7 @@ Two resolution methods are supported via the `mode` field:
   <network>/                # e.g., bitcoin/, regtest/, signet/
     api_key                # API key
     ldk-server.log         # Log file
+    ldk_node_postgres.lock  # Present when LDK Node state uses PostgreSQL
     ldk_node_data.sqlite   # LDK Node state (channels, on-chain wallet)
     ldk_server_data.sqlite # Payment and forwarding history
 ```
@@ -212,3 +237,7 @@ The mnemonic is the node's master secret, required to recover on-chain funds. On
 ldk-server generates a fresh 24-word BIP39 mnemonic at `<storage_dir>/keys_mnemonic` if the file
 does not already exist. `ldk_node_data.sqlite` holds channel state, both are required to recover
 channel funds. See [Operations - Backups](operations.md#backups) for backup guidance.
+
+When `[storage.postgres]` is configured, LDK Node wallet/channel state is stored in
+PostgreSQL instead of `ldk_node_data.sqlite`. The storage directory remains required for
+the mnemonic, API key, TLS material, logs, and `ldk_server_data.sqlite`.
--- a/ldk-server/src/util/config.rs
+++ b/ldk-server/src/util/config.rs
@@ -56,6 +56,7 @@ pub struct Config {
 	pub tls_config: Option<TlsConfig>,
 	pub grpc_service_addr: SocketAddr,
 	pub storage_dir_path: Option<String>,
+	pub ldk_node_storage: LdkNodeStorageConfig,
 	pub chain_source: ChainSource,
 	pub rgs_server_url: Option<String>,
 	pub lsps2_client_config: Option<LSPSClientConfig>,
@@ -92,6 +93,17 @@ pub struct TlsConfig {
 	pub hosts: Vec<String>,
 }
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum LdkNodeStorageConfig {
+	Sqlite,
+	Postgres {
+		connection_string: String,
+		db_name: Option<String>,
+		kv_table_name: Option<String>,
+		certificate_pem: Option<String>,
+	},
+}
+
 #[derive(Debug, PartialEq, Eq)]
 pub enum ChainSource {
 	Rpc {
@@ -126,6 +138,7 @@ struct ConfigBuilder {
 	tls_config: Option<TlsConfig>,
 	grpc_service_address: Option<String>,
 	storage_dir_path: Option<String>,
+	ldk_node_postgres: Option<PostgresStorageConfig>,
 	electrum_url: Option<String>,
 	esplora_url: Option<String>,
 	bitcoind_rpc_address: Option<String>,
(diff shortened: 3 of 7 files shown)


docs/configuration.md
```diff
@@ -212,3 +231,7 @@ The mnemonic is the node's master secret, required to recover on-chain funds. On
 ldk-server generates a fresh 24-word BIP39 mnemonic at `<storage_dir>/keys_mnemonic` if the file
 does not already exist. `ldk_node_data.sqlite` holds channel state, both are required to recover
 channel funds. See [Operations - Backups](operations.md#backups) for backup guidance.
+
+When `[storage.postgres]` is configured, LDK Node wallet/channel state is stored in
+PostgreSQL instead of `ldk_node_data.sqlite`. The storage directory remains required for
+the mnemonic, API key, TLS material, logs, and `ldk_server_data.sqlite`.
```
tankyleo, line 243: Would big routing nodes want `ldk_server_data.sqlite` to be in the postgres db too, given that they'll be forwarding a lot of payments ? Or is that for a later PR / out-of-scope here ?

tankyleo (commented):
skimmed briefly one question

docs/configuration.md
benthecarman: It will be after https://github.com/lightningdevkit/ldk-node/pull/772, just keeping the docs accurate of the local state of the repo. Will update once we merge and upgrade here

contrib/ldk-server-config.toml
```diff
@@ -28,7 +28,14 @@ alias = "ldk_server"                          # Lightning node alias
 
 # Storage settings
 [storage.disk]
-dir_path = "/tmp/ldk-server/"                 # Path for LDK and BDK data persistence, optional, defaults to ~/Library/Application Support/ldk-server/ on macOS, ~/.ldk-server/ on Linux
+dir_path = "/tmp/ldk-server/"                 # Path for local ldk-server data, optional, defaults to ~/Library/Application Support/ldk-server/ on macOS, ~/.ldk-server/ on Linux
+
+# Optional LDK Node PostgreSQL storage for wallet and channel state.
+#[storage.postgres]
+#connection_string = "postgresql://postgres:postgres@localhost:5432" # PostgreSQL connection string. Do not include dbname if db_name is set.
```
tankyleo, line 35: If it works I find the key value format to be cleaner, less error prone ie `host=localhost port=5432 user=postgres sslmode=disable`

docs/configuration.md
```diff
@@ -100,8 +100,27 @@ cooldown_secs = 3600
...
+```toml
+[storage.postgres]
+connection_string = "postgresql://postgres:postgres@localhost:5432"
+db_name = "ldk_db"
+kv_table_name = "ldk_data"
+certificate_path = "/path/to/postgres-ca.pem"
+```
+
+Only `connection_string` is required. `db_name`, `kv_table_name`, and `certificate_path`
+are optional. If `db_name` is set, do not also include a database name in the connection
```
tankyleo, line 121: Don't remember us explicitly rejecting a database name in the connection string anywhere in this patch ? Would it be good to add test coverage for this?

docs/configuration.md
```diff
@@ -100,8 +100,27 @@ cooldown_secs = 3600
...
+the node mnemonic, API key, TLS material, logs, and ldk-server's payment history when
+LDK Node wallet/channel state uses PostgreSQL.
+
+### `[storage.postgres]`
+
+Optional PostgreSQL storage for LDK Node wallet and channel state.
+
+```toml
+[storage.postgres]
+connection_string = "postgresql://postgres:postgres@localhost:5432"
```
tankyleo, line 114: Similar here, the key-value format would be better if we can have it.

ldk-server/src/util/config.rs
```diff
@@ -278,6 +295,27 @@ impl ConfigBuilder {
 			self.storage_dir_path = Some(storage_dir_path.clone());
 		}
 
+		if args.storage_postgres_connection_string.is_some()
+			|| args.storage_postgres_db_name.is_some()
+			|| args.storage_postgres_kv_table_name.is_some()
+			|| args.storage_postgres_certificate_path.is_some()
+		{
+			let mut postgres = self.ldk_node_postgres.take().unwrap_or_default();
```
tankyleo: Here we can do a `get_or_insert_default`, so we don't need the assignment at the end of the block.

ldk-server/src/util/config.rs
tankyleo: Similar here can do a `get_or_insert_default` if you care to :)

ldk-server/src/util/config.rs
```diff
@@ -499,6 +537,32 @@ impl ConfigBuilder {
 		let log_max_files = self.log_max_files.unwrap_or(DEFAULT_LOG_MAX_FILES);
 		let log_to_file = self.log_to_file.unwrap_or(true);
 
+		let ldk_node_storage = if let Some(postgres) = self.ldk_node_postgres {
+			LdkNodeStorageConfig::Postgres {
+				connection_string: postgres
+					.connection_string
+					.ok_or_else(|| missing_field_err("storage_postgres_connection_string"))?,
```
tankyleo, line 543: For this field I wonder if this is a sensible default:
`host=localhost port=5432 user=postgres sslmode=disable`

contrib/ldk-server-config.toml
benthecarman: tbh i always hate these, but can switch if we really want

docs/configuration.md
benthecarman: that's in ldk-node, added a test for this

ldk-server/src/util/config.rs
benthecarman: done

ldk-server/src/util/config.rs
benthecarman: done

ldk-server/src/util/config.rs
benthecarman: Imo probably better to not have a defualt connection string

joostjager:
Posting some AI comments that look sensible. Second one is in flight in https://github.com/lightningdevkit/ldk-node/pull/1000, but maybe it should be a prerequisite.

🤖 Requesting changes for two fund-safety blockers.

1. [The backend is selected only after reusing the existing mnemonic](https://github.com/lightningdevkit/ldk-server/blob/f25cb1d7eeab33c56a6483d8beb505e35d9fe5bd/ldk-server/src/main.rs#L252-L260), with no migration or continuity check. Adding `[storage.postgres]` to an existing SQLite deployment therefore preserves the same Lightning identity while an empty PostgreSQL target creates fresh wallet and channel-manager state and silently ignores `ldk_node_data.sqlite`. For a node with live channels, starting without its channel monitors can lose channel funds. Please require an explicit verified migration, persist and validate backend identity, or at minimum refuse this switch when local SQLite LDK state exists and the PostgreSQL destination is empty.

2. [The PostgreSQL build path](https://github.com/lightningdevkit/ldk-server/blob/f25cb1d7eeab33c56a6483d8beb505e35d9fe5bd/ldk-server/src/main.rs#L695-L711) does not acquire exclusive ownership of the database and table. Two deployments with the same mnemonic and target can both pass descriptor and network validation, read the same snapshot, and start the same node identity. The pinned PostgreSQL store uses process-local ordering locks and unconditional upserts, so overlapping instances can interleave stale channel-manager and monitor writes. Please hold a database advisory lock or equivalent lease for the node lifetime and fail the second writer before building the node.

The previous cross-network concern is fail-closed in the pinned dependency because the persisted BDK network is checked during wallet load. The target is still not network-isolated, however, so the operations claim that multiple networks can share one storage root without conflicts should state that PostgreSQL deployments need distinct database or table targets per network.

benthecarman:
1. Made it so we forbid migrating. As this is 0.1, i think this is fine for now and we can make a proper implementation in the future.
2. Made https://github.com/lightningdevkit/ldk-node/pull/1012

joostjager:
I’m not sure we should rush PostgreSQL support into ldk-server with an interim locking mechanism, only to replace it with a different approach later. That could also leave us with migration or compatibility concerns between deployed versions? https://github.com/lightningdevkit/ldk-node/pull/1012 does not look entirely trivial either.

benthecarman:
Now that #258 is merged with https://github.com/lightningdevkit/ldk-node/pull/1012 this should be good to go

tankyleo (commented):
Here's a review pass from codex. Perhaps the first item is out of scope. Also told codex user starting two instances with the same `network_dir` is out of scope.

  - [P1] Reject missing or changed PostgreSQL state on restart _ /home/ubuntu/ldk-server/ldk-server/src/main.rs:791-792
    When a node has already run on PostgreSQL and the configured database or table is later dropped, changed, or mistyped, postgres_lock_exists is true but this match accepts every PostgreSQL configuration as long as no SQLite file exists. build_with_postgres_store can then create an empty table and start fresh state under the same mnemonic and node ID, potentially endangering existing channels. Record and verify a backend identity or sentinel on subsequent PostgreSQL starts rather than only recording the backend type.

  - [P2] Bound the PostgreSQL certificate read _ /home/ubuntu/ldk-server/ldk-server/src/util/config.rs:558-561
    When certificate_path points to an oversized file or an unbounded pseudo-file such as /dev/zero, fs::read_to_string reads until EOF before PEM parsing, so startup can exhaust memory or hang. Other externally configured files use read_to_string_with_limit; apply an appropriate certificate-size limit here as well.

  - [P2] Keep default builds independent of system OpenSSL _ /home/ubuntu/ldk-server/ldk-server/Cargo.toml:16-16
    On Linux, enabling storage-postgres pulls in native-tls and openssl-sys without vendoring, so the default cargo build now fails unless pkg-config and OpenSSL development headers are installed, even when PostgreSQL is unused. This contradicts docs/getting-started.md stating that no other external dependencies are required; use the vendored TLS feature, make the backend opt-in, or document and install the new system prerequisites.

  - [P1] Filter PostgreSQL parameter values from debug logs _ /home/ubuntu/ldk-server/ldk-server/src/main.rs:1088-1088
    When [storage.postgres] is used with the default Debug log level, tokio_postgres::query logs every bound parameter, including the complete Vec<u8> passed to each persistence write. Passing this store to LDK Node therefore copies serialized payment records_including preimages_and channel/wallet state into the console or system journal and the server log file, while large state blobs can also rapidly consume log storage. Filter the tokio_postgres query targets or otherwise prevent driver parameter logging before using this store.

benthecarman:
> Here's a review pass from codex. Perhaps the first item is out of scope. Also told codex user starting two instances with the same `network_dir` is out of scope.
> 
>     * [P1] Reject missing or changed PostgreSQL state on restart _ /home/ubuntu/ldk-server/ldk-server/src/main.rs:791-792
>       When a node has already run on PostgreSQL and the configured database or table is later dropped, changed, or mistyped, postgres_lock_exists is true but this match accepts every PostgreSQL configuration as long as no SQLite file exists. build_with_postgres_store can then create an empty table and start fresh state under the same mnemonic and node ID, potentially endangering existing channels. Record and verify a backend identity or sentinel on subsequent PostgreSQL starts rather than only recording the backend type.

yeah agreed this is out of scope, seems overkill

>     * [P2] Bound the PostgreSQL certificate read _ /home/ubuntu/ldk-server/ldk-server/src/util/config.rs:558-561
>       When certificate_path points to an oversized file or an unbounded pseudo-file such as /dev/zero, fs::read_to_string reads until EOF before PEM parsing, so startup can exhaust memory or hang. Other externally configured files use read_to_string_with_limit; apply an appropriate certificate-size limit here as well.

fixed

>     * [P2] Keep default builds independent of system OpenSSL _ /home/ubuntu/ldk-server/ldk-server/Cargo.toml:16-16
>       On Linux, enabling storage-postgres pulls in native-tls and openssl-sys without vendoring, so the default cargo build now fails unless pkg-config and OpenSSL development headers are installed, even when PostgreSQL is unused. This contradicts docs/getting-started.md stating that no other external dependencies are required; use the vendored TLS feature, make the backend opt-in, or document and install the new system prerequisites.

this is an explicit choice, we want to use system tls because we don't want to have to ship an update for every openssl security update, user can update themselves. However, did remove the line in docs saying we don't need external deps.

>     * [P1] Filter PostgreSQL parameter values from debug logs _ /home/ubuntu/ldk-server/ldk-server/src/main.rs:1088-1088
>       When [storage.postgres] is used with the default Debug log level, tokio_postgres::query logs every bound parameter, including the complete Vec passed to each persistence write. Passing this store to LDK Node therefore copies serialized payment records_including preimages_and channel/wallet state into the console or system journal and the server log file, while large state blobs can also rapidly consume log storage. Filter the tokio_postgres query targets or otherwise prevent driver parameter logging before using this store.


discussed offline, seems overkill and weird to override user choice.

tankyleo (approved):

PR #246, opened 2026-07-17
Expose background probing service 

benthecarman:
Allow operators to enable either built-in probing strategy through the config file, CLI arguments, or environment variables. Validate strategy settings and document the risk of probes locking outbound liquidity.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -59,6 +59,45 @@ this node to go offline. Set `async_payments_role = "server"` to hold async paym
 and onion messages for peers. The server role requires an announceable node configuration.
 Leave the field unset to disable async payments.
 
+### `[probing]`
+
+Enables LDK Node's background probing service to train the payment scorer with current
+channel-liquidity information. Probing is disabled when this section and the corresponding
+CLI/environment options are absent.
+
+The `strategy` field selects one of two path-selection methods:
+
+- **`"high_degree"`** probes toward highly connected public nodes. The service uses the
+  payment scorer to select a path.
+- **`"random_walk"`** constructs random paths through the public graph. This strategy does
+  not use the payment scorer to select a path.
+
+The section accepts these fields:
+
+| Field | Requirement and default | Description |
+| --- | --- | --- |
+| `strategy` | Required. No default. | Selects `"high_degree"` or `"random_walk"`. Start with `"random_walk"` for a payment node. Use `"high_degree"` for a dedicated probing node. |
+| `top_node_count` | Required for `"high_degree"`. No default. Start with `100`. | Sets the number of highly connected public nodes in the destination set. The strategy cycles through this set. The value must be greater than `0`. |
+| `max_hops` | Required for `"random_walk"`. No default. Start with `5`. | Sets the maximum number of hops in a random path. The value must be at least `2`. LDK Node changes values greater than `19` to `19`. |
+| `interval_secs` | Optional. The default is `10`. | Sets the number of seconds between probe attempts. LDK Node changes `0` to its minimum interval of 100 milliseconds. |
+| `max_locked_msat` | Optional. The default is `100000000` (100,000 satoshis). | Limits the total amount and pending fees of in-flight probes. The service skips a probe that exceeds the remaining limit. Built-in probes use 1,000,000 through 10,000,000 millisatoshis before routing fees. |
+| `diversity_penalty_msat` | Optional. The default is `0`. | Adds a virtual routing cost to recently probed channels. The service does not pay this amount. The cost decreases to zero over 24 hours and encourages different paths. This field only affects `"high_degree"`. |
+| `cooldown_secs` | Optional. The default is `3600`. | Sets the time before `"high_degree"` can select the same destination again. The strategy starts a new cycle immediately after it probes all destinations. |
+
+```toml
+[probing]
+strategy = "high_degree"
+top_node_count = 100
+interval_secs = 30
+max_locked_msat = 100000000
+diversity_penalty_msat = 250
+cooldown_secs = 3600
+```
+
+> [!CAUTION]
+> Probes send real HTLCs over real channels. A probe can lock outbound liquidity
+> until its HTLC expires. Use `max_locked_msat` to limit this risk.
+
 ### `[storage.disk]`
 
 Where persistent data is stored. Defaults to `~/.ldk-server/` on Linux and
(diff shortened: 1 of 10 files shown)


docs/configuration.md
```diff
@@ -59,6 +59,37 @@ this node to go offline. Set `async_payments_role = "server"` to hold async paym
...
+- **`"random_walk"`** - constructs random paths through the public graph. Set `max_hops`
+  to limit path length; values below `2` are clamped to `2`.
+
+```toml
+[probing]
+strategy = "high_degree"
+top_node_count = 100
+interval_secs = 30
+max_locked_msat = 500000
+diversity_penalty_msat = 250
```
tankyleo, line 93: Can we add a few quick words on what this is somewhere ? Either in `contrib/ldk-server-config.toml` or here, I don't know what this is.

docs/configuration.md
```diff
@@ -59,6 +59,37 @@ this node to go offline. Set `async_payments_role = "server"` to hold async paym
...
+  `top_node_count` to control how many top nodes are cycled through.
+- **`"random_walk"`** - constructs random paths through the public graph. Set `max_hops`
+  to limit path length; values below `2` are clamped to `2`.
+
+```toml
+[probing]
+strategy = "high_degree"
+top_node_count = 100
+interval_secs = 30
+max_locked_msat = 500000
```
tankyleo: Codex:
```
  - [P2] Raise the documented probing budget _ /home/ubuntu/ldk-server/docs/configuration.md:80-80
    With this value, users who copy the example get a prober that never sends probes: LDK Node's built-in strategies choose probe amounts starting at 1,000,000 msat, and the prober skips any path whose amount would exceed max_locked_msat. Use at least the minimum probe amount so the documented
    configuration actually trains the scorer.
```

docs/configuration.md
benthecarman: added

e2e-tests/tests/e2e.rs
```diff
@@ -1237,12 +1200,12 @@ async fn test_forwarded_payment_event() {
...
 	let mut builder_c = ldk_node::Builder::from_config(config_c);
 	let (rpc_host, rpc_port, rpc_user, rpc_password) = bitcoind.rpc_details();
-	builder_c.set_chain_source_bitcoind_rpc(rpc_host, rpc_port, rpc_user, rpc_password);
+	builder_c.set_chain_source_bitcoind_rpc(rpc_host, rpc_port, rpc_user, rpc_password, None);
 
 	// Set B as LSPS2 LSP for C
 	let b_node_id = ldk_node::bitcoin::secp256k1::PublicKey::from_str(server_b.node_id()).unwrap();
 	let b_addr = SocketAddress::from_str(&format!("127.0.0.1:{}", server_b.p2p_port)).unwrap();
-	builder_c.set_liquidity_source_lsps2(b_node_id, b_addr, None);
+	builder_c.add_liquidity_source(b_node_id, b_addr, None, true);
```
tankyleo, line 1208: Should this be false since we actually confirm the channel further below?

docs/configuration.md
```diff
@@ -59,6 +59,44 @@ this node to go offline. Set `async_payments_role = "server"` to hold async paym
...
+[probing]
+strategy = "high_degree"
+top_node_count = 100
+interval_secs = 30
+max_locked_msat = 100000000
+diversity_penalty_msat = 250
+cooldown_secs = 3600
+```
+
+> **Caution:** Probes send real HTLCs over real channels. A probe can lock outbound liquidity
```
tankyleo: I think we need this to be Warning instead of Caution to have the little yellow sidebar display from Github ?

docs/configuration.md
```diff
@@ -59,6 +59,44 @@ this node to go offline. Set `async_payments_role = "server"` to hold async paym
...
+- **`"random_walk"`** constructs random paths through the public graph. This strategy does
+  not use the payment scorer to select a path.
+
+The section accepts these fields:
+
+| Field | Requirement and default | Description |
+| --- | --- | --- |
+| `strategy` | Required. | Selects `"high_degree"` or `"random_walk"`. |
+| `top_node_count` | Required for `"high_degree"`. | Sets the number of highly connected public nodes in the destination set. The strategy cycles through this set. The value must be greater than `0`. |
+| `max_hops` | Required for `"random_walk"`. | Sets the maximum number of hops in a random path. LDK Node clamps the value to the range `2` through `19`. |
```
tankyleo: Should we tell people the default value for these three lines in the "Requirement and default" column ? Or maybe this is done in the example table below

ldk-server/src/util/config.rs
```diff
@@ -766,6 +811,76 @@ fn parse_async_payments_role(role: &str) -> io::Result<AsyncPaymentsRole> {
...
+			ProbingConfigBuilder::high_degree(top_node_count)
+		},
+		"random_walk" | "random-walk" => {
+			if top_node_count.is_some() {
+				return Err(io::Error::new(
+					io::ErrorKind::InvalidInput,
+					"`probing.top_node_count` only applies to the `high_degree` strategy",
+				));
+			}
+			let max_hops = max_hops.ok_or_else(|| missing_field_err("probing.max_hops"))?;
```
tankyleo, line 854: We reject `top_node_count == 0`, so should we also reject `max_hops <=1` ? Especially since noone will read the documentation about the clamping to the value of 2 ?

docs/configuration.md
benthecarman: Warning gives the requested yellow GitHub alert, Caution renders red

e2e-tests/tests/e2e.rs
benthecarman: We do mine the blocks but the LSP uses `client_trusts_lsp = true` so it never brodcasts the initial funding tx until we receive a payment.

ldk-server/src/util/config.rs
benthecarman: added

docs/configuration.md
benthecarman: there isn't a default but added some recommendations

tankyleo:
CI is failing is that a known failure ?

docs/configuration.md
tankyleo: ah yes, still needs some small tweak i believe, it's not rendering for me.

https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts

benthecarman:
> CI is failing is that a known failure ?

was a flake with a old bitreq version according to codex, updated the cargo.lock to latest

tankyleo (approved):

PR #250, opened 2026-07-27
Expose zero-fee commitment channel configuration

tankyleo:
This setting was recently introduced in LDK-node, we wire it to the different configuration interfaces of LDK-server.

Co-Authored-By: Codex

### DIFF
--- a/ldk-server/src/util/config.rs
+++ b/ldk-server/src/util/config.rs
@@ -70,6 +70,7 @@ pub struct Config {
 	pub pathfinding_scores_source_url: Option<String>,
 	pub probing_config: Option<ProbingConfig>,
 	pub async_payments_role: Option<AsyncPaymentsRole>,
+	pub enable_zero_fee_commitments: bool,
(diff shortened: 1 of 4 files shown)


ldk-server/src/util/config.rs
```diff
@@ -1731,7 +1753,8 @@ mod tests {
 			"#;
 
 		fs::write(storage_path.join(config_file_name), toml_config).unwrap();
-		assert!(load_config(&args_config).is_ok());
+		let config = load_config(&args_config).unwrap();
+		assert!(!config.enable_zero_fee_commitments);
```
benthecarman: this line doesn't really make sense in this test, we don't do any other asserts like this

benthecarman (commented):
small nit, otherwise lgtm

ldk-server/src/util/config.rs
tankyleo: reverted the chunk below

benthecarman (approved):

PR #253, opened 2026-08-03
Make mnemonic RNG selection explicit

benthecarman:
Use direct OS randomness for the server's one-time mnemonic generation and an explicit OsRng in the E2E setup. A thread-local generator is not needed for these infrequent operations.

### DIFF
--- a/ldk-server/src/util/entropy.rs
+++ b/ldk-server/src/util/entropy.rs
@@ -12,12 +12,13 @@ use std::str::FromStr;
 use std::{fs, io};
 
 use ldk_node::bip39::Mnemonic;
-use ldk_node::entropy::{generate_entropy_mnemonic, NodeEntropy};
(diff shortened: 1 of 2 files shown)


ldk-server/src/util/entropy.rs
```diff
@@ -46,6 +47,12 @@ pub(crate) fn load_or_generate_node_entropy(storage_dir: &Path) -> io::Result<No
 	Ok(NodeEntropy::from_bip39_mnemonic(mnemonic, None))
 }
 
+fn generate_entropy_mnemonic() -> io::Result<Mnemonic> {
```
tankyleo, line 50: I think this deserves a fix in `ldk-node` and `bip39` no ?

benthecarman: yeah i guess that makes more sense

PR #255, opened 2026-08-10
Update MCP to 2026-07-28 protocol

benthecarman:
Add stateless discovery and per-request protocol negotiation. Return the required result metadata and cache hints while retaining compatibility with the legacy initialization flow. 

Keeps legacy support for the old protocol for now, unsure if this is the correct route to go if we should only support the latest.


benthecarman:
Closed in favor of #256

PR #256, opened 2026-08-10
Update to MCP 2026-07-28 without backwards compat.

tnull:
Alternative to #255 (which I only saw after opening this).

Here we don't keep the old version around, just go for the new thing.

(cc @benthecarman let me know what you prefer).

### DIFF
--- a/ldk-server-mcp/src/main.rs
+++ b/ldk-server-mcp/src/main.rs
@@ -14,15 +14,83 @@ mod tools;
 
 use ldk_server_client::client::LdkServerClient;
 use ldk_server_client::ldk_server_grpc::api::GetNodeInfoRequest;
+use serde::Serialize;
 use serde_json::Value;
 use tokio::io::{AsyncBufReadExt, AsyncWriteExt, BufReader};
 
-use crate::mcp::InitializeResult;
+use crate::mcp::{request_protocol_version, DiscoverResult, ListToolsResult, PROTOCOL_VERSION};
 use crate::protocol::{
 	JsonRpcErrorResponse, JsonRpcRequest, JsonRpcResponse, INVALID_PARAMS, METHOD_NOT_FOUND,
-	PARSE_ERROR,
+	PARSE_ERROR, UNSUPPORTED_PROTOCOL_VERSION,
 };
-use crate::tools::build_tool_registry;
+use crate::tools::{build_tool_registry, ToolRegistry};
+
+fn result_response(id: Value, result: impl Serialize) -> Value {
+	serde_json::to_value(JsonRpcResponse::new(id, serde_json::to_value(result).unwrap())).unwrap()
+}
+
+fn error_response(id: Value, code: i64, message: impl Into<String>) -> Value {
+	serde_json::to_value(JsonRpcErrorResponse::new(id, code, message.into())).unwrap()
+}
+
+fn error_response_with_data(
+	id: Value, code: i64, message: impl Into<String>, data: Value,
+) -> Value {
+	serde_json::to_value(JsonRpcErrorResponse::with_data(id, code, message.into(), data)).unwrap()
+}
+
+async fn handle_request(
+	request: JsonRpcRequest, client: &LdkServerClient, registry: &ToolRegistry,
+) -> Value {
+	let id = request.id.clone();
+	if request.method == "initialize" {
+		return error_response_with_data(
+			id,
+			METHOD_NOT_FOUND,
+			format!("initialize is not supported; this server requires MCP {PROTOCOL_VERSION}"),
+			serde_json::json!({ "supported": [PROTOCOL_VERSION] }),
+		);
+	}
+
+	let protocol_version = match request_protocol_version(request.params.as_ref()) {
+		Ok(protocol_version) => protocol_version,
+		Err(message) => return error_response(id, INVALID_PARAMS, message),
+	};
+	if protocol_version != PROTOCOL_VERSION {
+		return error_response_with_data(
+			id,
+			UNSUPPORTED_PROTOCOL_VERSION,
+			"Unsupported protocol version",
+			serde_json::json!({
+				"supported": [PROTOCOL_VERSION],
+				"requested": protocol_version,
+			}),
+		);
+	}
+
+	match request.method.as_str() {
+		"server/discover" => result_response(id, DiscoverResult::new()),
+		"tools/list" => result_response(id, ListToolsResult::new(registry.list_tools())),
+		"tools/call" => {
+			let params = request.params.as_ref().unwrap();
+			let Some(tool_name) = params.get("name").and_then(Value::as_str) else {
+				return error_response(id, INVALID_PARAMS, "Missing required parameter: name");
+			};
+			let tool_args = match params.get("arguments") {
+				Some(arguments) if !arguments.is_object() => {
+					return error_response(id, INVALID_PARAMS, "arguments must be an object");
+				},
+				Some(arguments) => arguments.clone(),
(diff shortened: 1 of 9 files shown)


benthecarman:
Yeah based on the goose team's comments this seems like the better way to go

ldk-server-mcp/src/main.rs
```diff
@@ -14,15 +14,83 @@ mod tools;
...
+use crate::mcp::{request_protocol_version, DiscoverResult, ListToolsResult, PROTOCOL_VERSION};
 use crate::protocol::{
 	JsonRpcErrorResponse, JsonRpcRequest, JsonRpcResponse, INVALID_PARAMS, METHOD_NOT_FOUND,
-	PARSE_ERROR,
+	PARSE_ERROR, UNSUPPORTED_PROTOCOL_VERSION,
 };
-use crate::tools::build_tool_registry;
+use crate::tools::{build_tool_registry, ToolRegistry};
+
+fn result_response(id: Value, result: impl Serialize) -> Value {
```
benthecarman: can we put all these functions under `fn main` instead of above it

ldk-server-mcp/src/main.rs
```diff
@@ -14,15 +14,83 @@ mod tools;
...
+fn error_response_with_data(
+	id: Value, code: i64, message: impl Into<String>, data: Value,
+) -> Value {
+	serde_json::to_value(JsonRpcErrorResponse::with_data(id, code, message.into(), data)).unwrap()
+}
+
+async fn handle_request(
+	request: JsonRpcRequest, client: &LdkServerClient, registry: &ToolRegistry,
+) -> Value {
+	let id = request.id.clone();
```
benthecarman: i dont think we need this clone

ldk-server-mcp/tests/integration.rs
```diff
@@ -11,6 +11,7 @@ use std::io::{BufRead, BufReader, Write};
 
 use serde_json::{json, Value};
 
+const PROTOCOL_VERSION: &str = "2026-07-28";
```
benthecarman: we have this in mcp.rs

ldk-server-mcp/src/main.rs
```diff
@@ -14,15 +14,83 @@ mod tools;
...
+	serde_json::to_value(JsonRpcErrorResponse::new(id, code, message.into())).unwrap()
+}
+
+fn error_response_with_data(
+	id: Value, code: i64, message: impl Into<String>, data: Value,
+) -> Value {
+	serde_json::to_value(JsonRpcErrorResponse::with_data(id, code, message.into(), data)).unwrap()
+}
+
+async fn handle_request(
```
benthecarman: claude's main concern:

**Validation ordering in `handle_request` is the root cause of the three most important findings.** The full `_meta` shape check runs before the protocol-version comparison and before method dispatch. As a result, a client on a different protocol version can get a bare `-32602` about `clientCapabilities` instead of the `-32022` negotiation error with `data.supported`, and unknown methods without `_meta` get `-32602` instead of `METHOD_NOT_FOUND`. A real legacy `ping` (no params) hits the same path and gets a misleading “params must be an object” error; the test that claims to cover this passes only because the harness injects a valid `_meta` into every request.

ldk-server-mcp/src/main.rs
tnull: Done in fixup

ldk-server-mcp/src/main.rs
tnull: Done in fixup

ldk-server-mcp/tests/integration.rs
tnull: Done in fixup

ldk-server-mcp/src/main.rs
tnull: Done in fixup

ldk-server-mcp/src/tools/mod.rs
```diff
@@ -52,17 +52,21 @@ impl ToolRegistry {
...
-				let text = serde_json::to_string(&value)
-					.unwrap_or_else(|e| format!("Failed to serialize response: {e}"));
-				ToolCallResult::success(text)
+				let text = serde_json::to_string(&value).map_err(|e| {
+					McpError::internal(format!("Failed to serialize response: {e}"))
+				})?;
+				Ok(ToolCallResult::success(text))
 			},
-			Err(e) => ToolCallResult::error(format!("{}: {}", e.category(), e.message)),
+			Err(e) if e.is_tool_execution() => {
```
benthecarman, line 66: **Argument-validation errors now violate the spec** (`tools/mod.rs:66`). I earlier relayed that converting invalid-argument errors to JSON-RPC `-32602` was spec-prescribed — that was wrong. The spec’s tools Error Handling section explicitly lists input validation errors (“value out of range”, wrong format) under _Tool Execution Errors_ reported with `isError: true`, reserving protocol errors for unknown tools and schema-shape failures. The PR moves all 37 tools’ argument parsing to hard `-32602` envelope errors, which breaks the LLM self-correction loop the spec designed `isError` for. This is the most consequential finding after the notification bug.

ldk-server-mcp/src/main.rs
```diff
@@ -93,74 +97,92 @@ async fn main() {
...
+	match method.as_str() {
+		"server/discover" => result_response(id, DiscoverResult::new()),
+		"tools/list" => result_response(id, ListToolsResult::new(registry.list_tools())),
+		"tools/call" => {
+			let params = params.as_ref().unwrap();
+			let Some(tool_name) = params.get("name").and_then(Value::as_str) else {
+				return error_response(id, INVALID_PARAMS, "Missing required parameter: name");
+			};
+			let tool_args = match params.get("arguments") {
+				Some(arguments) if !arguments.is_object() => {
```
benthecarman, line 175: **`"arguments": null` is rejected while an omitted key succeeds** (`main.rs:175`). Confirmed by running the binary — many client serializers emit `null` for “no arguments,” and the old code tolerated it.

ldk-server-mcp/src/protocol.rs
```diff
@@ -52,19 +65,56 @@ impl From<LdkServerError> for McpError {
...
+		let Some(message) = value.as_object() else {
+			return Err(JsonRpcErrorResponse::new(
+				Value::Null,
+				INVALID_REQUEST,
+				"Invalid Request".to_string(),
+			));
+		};
+
+		let id = message.get("id").cloned();
+		let response_id = id
```
benthecarman, line 90: **`ldk-server-mcp/src/protocol.rs` — `JsonRpcRequest::from_value`:** the `!valid` early return runs before the `let Some(id) = id else { return Ok(None) }` notification bail-out, so a malformed notification (e.g. `{"jsonrpc":"2.0","method":"notifications/cancelled","params":null}`) gets an `id: null` error response on stdout. JSON-RPC 2.0 and MCP both say notifications MUST NOT receive a response, and this crate’s [CLAUDE.md](http://claude.md/) says they’re ignored. Also, an explicit `"id": null` now maps to `Some(Value::Null)` and gets rejected, where serde previously folded it to `None` (silent).

Fix: move the notification bail-out above the validity check and treat a `null` id as absent, e.g. `let Some(id) = id.filter(|id| !id.is_null()) else { return Ok(None); };` — then simplify `response_id` and the id clause of `valid`, which can assume a present, non-null id. Please also extend `test_notification_no_response` with a `params: null` (or array-params) notification asserting no output line.

benthecarman (commented):
overall looks good to me, just some final review comments on correctness from claude

PR #258, opened 2026-08-18
Update ldk-node dependency & expose bolt12 proofs

benthecarman:
Update ldk-node and adapt payment events to its current API. Expose payment IDs in events and retain manual handling for unknown BOLT 11 payments.

Add BOLT 12 payer-proof creation to the gRPC, CLI, and MCP interfaces. Include the preimage and invoice in successful-payment events for stateless proof creation.

Use ldk-node pagination for payment history. Remove duplicate payment records from the ldk-server SQLite store, which now contains only forwarded-payment history.


ldk-server/src/api/mod.rs
```diff
@@ -132,6 +136,31 @@ pub(crate) fn proto_to_node_custom_tlv(proto: &ProtoCustomTlvRecord) -> NodeCust
 	NodeCustomTlvRecord { type_num: proto.type_num, value: proto.value.to_vec() }
 }
 
+/// Resolve the local payment id for a BOLT11 payment hash.
+///
+/// The claim/fail APIs now take a payment id. Pending inbound matches are preferred so a
+/// later retry of the same hash does not claim an already-settled record.
+pub(crate) fn bolt11_payment_id_for_hash(
```
tnull: I don't think we should do this. Iterating all payments will be prohibitively expensive, esp. after only keeping a cache of them in memory. In LDK Node we decided against keeping a reverse index. I think this PR simply has to update the API to also go with `fail_for_id`/`claim_for_id` rather than maintaining the old API shape.

benthecarman: oh woops, had that locally but lost when rebasing i think, pushed as a fixup

benthecarman:
Rebased and updated ldk-node to new commit with the pagination changes. Now using the paginated payments instead of the ldk-server version

ldk-server/src/api/error.rs
```diff
@@ -89,8 +89,10 @@ impl From<NodeError> for LdkServerError {
 			| NodeError::InvalidNodeAlias
 			| NodeError::InvalidDateTime
 			| NodeError::InvalidFeeRate
+			| NodeError::InvalidPageToken
```
tnull, line 93: It seems we're not returning that?

Codex

>   - P2 — Return InvalidRequestError for invalid page tokens. /home/tnull/workspace/ldk-server-pr-258/ldk-server/src/api/list_payments.rs:25 maps every Node::list_payments failure to InternalServerError, including the newly added NodeError::InvalidPageToken. Using the existing From<NodeError>
>     conversion would correctly classify malformed client tokens.

tnull (commented):
Looks good, but we should probably also include the new channel type in `list_channels` now.

ldk-server/src/api/error.rs
benthecarman: fixed

ldk-server/src/api/error.rs
```diff
@@ -122,6 +125,7 @@ impl From<NodeError> for LdkServerError {
 			| NodeError::GossipUpdateTimeout
 			| NodeError::LiquiditySourceUnavailable
 			| NodeError::LiquidityRequestFailed
+			| NodeError::PayerProofCreationFailed
```
tankyleo: Codex wants to classify this to an `InvalidRequestError` rather than an `InternalServerError`. It's not perfect, but seems `InvalidRequestError` is more likely if we hit `PayerProofCreationFailed`

benthecarman: Yeah not perfect mapping, but fixed

ldk-server/src/util/metrics.rs
```diff
@@ -31,7 +31,36 @@
...
+
+#[derive(Default)]
+struct PaymentCounts {
+	total: i64,
+	successful: i64,
+	pending: i64,
+	failed: i64,
+}
+
+fn payment_status_counts(node: &Node) -> Result<PaymentCounts, NodeError> {
```
tankyleo, line 40: From codex, your call if you want to address this here:
>   - [P2] ldk-server/src/util/metrics.rs:45 scans and deserializes the entire payment history every metrics interval. The new LDK Node store is bounded and paginated, so this becomes repeated database work that grows without
    bound. It also runs twice at startup because ldk-server/src/main.rs:338 initializes metrics and Tokio’s first interval tick at line 342 is immediate. At minimum, delay the first poll; longer-term, maintain counters
    incrementally or expose counts from ldk-node.

ldk-server/src/main.rs
```diff
@@ -693,22 +701,25 @@ fn main() {
 	log::logger().flush();
 }
 
-fn send_event_and_upsert_payment(
+fn send_payment_event(
```
tankyleo, line 760: Codex: 
>   - [P3] ldk-server/src/main.rs:704 retrieves and converts payment details even when there are no subscribers, then deep-clones the resulting protobuf Payment. PaymentSuccessful additionally serializes and hex-encodes the
    invoice beforehand at ldk-server/src/main.rs:523. A receiver_count() == 0 fast path could acknowledge the event immediately, and the callback could take Payment by value to eliminate the clone.

docs/api-guide.md
tankyleo, line 1: From codex:
>   - The payer-proof feature deserves a short lifecycle section in docs/api-guide.md:198: subscribe before paying, retain all three PaymentSuccessful fields, and note that events are not replayed and static-invoice payments
    cannot produce proofs.
 
>   - The pagination guide at docs/api-guide.md:248 refers to a number_of_payments request field that does not exist in the gRPC API. That is a CLI-side aggregation option; the RPC’s page size is backend-controlled.

> - Document duplicate and late hodl-payment handling — docs/api-guide.md:233-242
    The lifecycle omits ldk-node’s safety requirements: callers must never reuse a payment hash, and one invoice may be paid multiple times, producing separately actionable PaymentClaimable events. Claiming each event accepts
    each payment, so unexpected duplicate or late events must be explicitly failed rather than ignored or claimed. Add this guidance here, and mirror it in api.proto on Bolt11ReceiveForHashRequest and PaymentClaimable so
    generated API users see it: use a unique hash per invoice, handle every event by payment_id, claim only expected payments before claim_deadline, and call Bolt11FailForId for unexpected duplicate or late payments.

tankyleo (commented):
I asked Codex to cast a broader net, here's what it found, feel free to dismiss aggressively, but all seemed worth taking a look to me.

tankyleo:
Also confirmed this patch now keeps my fans quiet on mainnet

ldk-server/src/main.rs
```diff
@@ -491,77 +495,83 @@ fn main() {
...
+										bolt12_invoice,
+									})
+								},
 								&event_node,
-								&event_sender,
-								Arc::clone(&paginated_store));
+								&event_sender);
 
 							if let Some(metrics) = &metrics {
 								metrics.update_payments_count(true);
```
tankyleo: Codex found this can race with `update_all_pollable_metrics` and suggests to remove `fn update_payments_count` entirely and have `update_all_pollable_metrics` be the sole writer.

I then asked about the other metrics here's what it had to say:
```
_ Yes, but not all in the same way.

  - Payment counters had the strongest race: update_all_pollable_metrics() performed an absolute store(), while update_payments_count() performed fetch_add() on the same value. That could permanently double-count or lose an
    update until reconciliation.

  - Balance gauges can still race: both the polling task and payment event handlers call update_all_balances(). Because both write absolute snapshots, an older, slower call could overwrite a newer snapshot. This causes
    temporary staleness, not double-counting, and the next refresh corrects it.

  - Channel metrics avoid this particular race: total_channels_count is event-driven, while public/private channel counts are poll-driven. They do not concurrently update the same atomic value.
  - Peer count is polling-only.

  For complete consistency, balance refreshes should also be serialized or assigned to a single updater. Prometheus gauges usually tolerate brief staleness, whereas payment counters decreasing or double-counting is more
  operationally problematic.
```

ldk-server/src/util/metrics.rs
```diff
@@ -92,20 +128,10 @@ impl Metrics {
...
+		match payment_status_counts(node) {
+			Ok(counts) => self.store_payment_counts(counts),
+			Err(e) => error!("Failed to initialize payment metrics: {e}"),
 		}
-		self.total_successful_payments_count.store(successful_payments_count, Ordering::Relaxed);
-		self.total_failed_payments_count.store(failed_payments_count, Ordering::Relaxed);
-		self.total_pending_payments_count.store(pending_payments_count, Ordering::Relaxed);
 
 		let channels_count = node.list_channels().len() as i64;
 		self.total_channels_count.store(channels_count, Ordering::Relaxed);
```
tankyleo: Codex suggests we initialize other metrics in addition to the total channels count:
```diff

    Initialize poll-only metrics at startup

    Seed peer and channel visibility gauges before the first delayed metrics poll.\n\nAI assistance: OpenAI Codex was used for this change.

diff --git a/ldk-server/src/util/metrics.rs b/ldk-server/src/util/metrics.rs
index 1c92b87..893b1b9 100644
--- a/ldk-server/src/util/metrics.rs
+++ b/ldk-server/src/util/metrics.rs
@@ -133,9 +133,20 @@ impl Metrics {
                        Err(e) => error!("Failed to initialize payment metrics: {e}"),
                }

-               let channels_count = node.list_channels().len() as i64;
-               self.total_channels_count.store(channels_count, Ordering::Relaxed);
+               let all_channels = node.list_channels();
+               self.total_channels_count.store(all_channels.len() as i64, Ordering::Relaxed);
+
+               let public_channels_count =
+                       all_channels.iter().filter(|channel_details| channel_details.is_announced).count()
+                               as i64;
+               self.total_public_channels_count.store(public_channels_count, Ordering::Relaxed);

+               let private_channels_count =
+                       all_channels.iter().filter(|channel_details| !channel_details.is_announced).count()
+                               as i64;
+               self.total_private_channels_count.store(private_channels_count, Ordering::Relaxed);
+
+               self.update_peer_count(node);
                self.update_all_balances(node);
        }
```

ldk-server/src/main.rs
```diff
@@ -695,22 +705,33 @@ fn main() {
...
+			if let Err(e) = event_sender.send(EventEnvelope { event: Some(event) }) {
+				debug!("No event subscribers connected, skipping event: {e}");
+			}
+
+			if let Err(e) = event_node.event_handled() {
+				error!("Failed to mark event as handled: {e}");
+			}
+		},
+		Ok(None) => error!("Unable to find payment with payment ID: {payment_id}"),
+		Err(e) => error!("Failed to retrieve payment with payment ID {payment_id}: {e}"),
```
tankyleo: Grok 4.6 found this: should we call event_node.event_handled() in these branches too ?
```
[bug] ldk-server/src/main.rs:728 _ send_payment_event skips event_handled() when payment-store lookup fails, which can stall the event queue
```
Then codex:
```
[P1] Resolve payment events when payment lookup fails _
With an active subscriber, Ok(None) and Err(_) return without calling event_handled().
Since LDK Node repeatedly returns the queue-head event until acknowledged,
a missing or unreadable payment record blocks every later event and creates a tight retry loop.
Emit a partial event with payment: None, or otherwise explicitly acknowledge/drop the event;
use controlled backoff if retrieval errors should be retried.
```

tankyleo: this is the suggested patch:
```diff
diff --git a/ldk-server/src/main.rs b/ldk-server/src/main.rs
index c5fd800..e1fab26 100644
--- a/ldk-server/src/main.rs
+++ b/ldk-server/src/main.rs
@@ -706,27 +706,23 @@ fn send_payment_event(
 ) {
        if event_sender.receiver_count() == 0 {
                debug!("No event subscribers connected, skipping payment event");
-               if let Err(e) = event_node.event_handled() {
-                       error!("Failed to mark event as handled: {e}");
+       } else {
+               match event_node.payment(payment_id) {
+                       Ok(Some(payment_details)) => {
+                               let payment = payment_to_proto(payment_details);
+
+                               let event = payment_to_event(payment);
+                               if let Err(e) = event_sender.send(EventEnvelope { event: Some(event) }) {
+                                       debug!("No event subscribers connected, skipping event: {e}");
+                               }
+                       },
+                       Ok(None) => error!("Unable to find payment with payment ID: {payment_id}"),
+                       Err(e) => error!("Failed to retrieve payment with payment ID {payment_id}: {e}"),
                }
-               return;
        }

-       match event_node.payment(payment_id) {
-               Ok(Some(payment_details)) => {
-                       let payment = payment_to_proto(payment_details);
-
-                       let event = payment_to_event(payment);
-                       if let Err(e) = event_sender.send(EventEnvelope { event: Some(event) }) {
-                               debug!("No event subscribers connected, skipping event: {e}");
-                       }
-
-                       if let Err(e) = event_node.event_handled() {
-                               error!("Failed to mark event as handled: {e}");
-                       }
-               },
-               Ok(None) => error!("Unable to find payment with payment ID: {payment_id}"),
-               Err(e) => error!("Failed to retrieve payment with payment ID {payment_id}: {e}"),
+       if let Err(e) = event_node.event_handled() {
+               error!("Failed to mark event as handled: {e}");
        }
 }
```

ldk-server/src/main.rs
benthecarman: thanks fixed

ldk-server/src/util/metrics.rs
benthecarman: fixed, collect all before a delayed poll

ldk-server/src/main.rs
benthecarman: done

benthecarman:
Also fixed the proto definition to be a single string for the `PageToken` and it's handling in the cli.

docs/api-guide.md
tankyleo: While chatting about how we call `event_handled` even if we hit a persistence read failure and fail to send the payment event in `fn send_payment_event` instead of not handling the event and replaying it hoping to forward it later, I eventually landed on this documentation update below. TLDR people should be aware notifications are currently best effort.
```diff
diff --git a/docs/api-guide.md b/docs/api-guide.md
index fb0e737..50cc052 100644
--- a/docs/api-guide.md
+++ b/docs/api-guide.md
@@ -216,8 +216,21 @@ See [Pagination](#pagination) below for how to page through results.
 | `SpliceNegotiated` | A channel splice was negotiated and the funding transaction is pending confirmation |
 | `SpliceNegotiationFailed` | A channel splice negotiation round failed                       |

-Events are broadcast to all connected subscribers. The server uses a bounded broadcast channel
-(capacity 1024). A slow subscriber that falls behind will miss events.
+> [!WARNING]
+> `SubscribeEvents` is a best-effort live notification stream. Events are not persisted for
+> subscribers, cannot be replayed after reconnecting, and have no client acknowledgement. Receipt
+> by the server's broadcast channel does not guarantee that a client received or processed an
+> event.
+
+Events are broadcast to all currently connected subscribers. The server uses a bounded broadcast
+channel (capacity 1024), so a slow subscriber that falls behind will miss events. Disconnected
+clients also miss events and receive only new events after reconnecting. If the server cannot read
+data required to construct an event, it logs the error and may skip that event so the event queue
+can continue processing.
+
+Treat events as notifications rather than authoritative history. After reconnecting, reconcile
+recoverable state with APIs such as `GetPaymentDetails`, `ListPayments`, `ListForwardedPayments`,
+and `ListChannels`. Some event-only fields cannot be recovered through these APIs.
```

docs/api-guide.md
```diff
@@ -229,27 +230,45 @@ GET /metrics
...
+4. **Decide before `claim_deadline`:**
+    - **Accept an expected payment:** Call `Bolt11ClaimForId` with its payment ID and preimage.
+    - **Reject an unexpected payment:** Call `Bolt11FailForId` with its payment ID. Reject duplicate
+      and late payments instead of ignoring or claiming them.
 
 The payment is held in a pending state until you explicitly claim or fail it. **You must
-always call one of these.** If you do neither, the HTLC will eventually time out, which
-can cause a force-closure of the channel.
+always handle each event.** If you do not, the HTLC will eventually time out. This can cause a
+force-closure of the channel.
```
tankyleo: Somewhat related to the above discussion, also landed on this update here:
```diff
-The payment is held in a pending state until you explicitly claim or fail it. **You must
-always handle each event.** If you do not, the HTLC will eventually time out. This can cause a
-force-closure of the channel.
+The payment is held in a pending state until you claim it, fail it, or its `claim_deadline` is
+reached. `PaymentClaimable` notifications are best-effort and are not replayed. If you miss the
+event or do not act before the deadline, LDK Node automatically fails the HTLC backward and the
+payment can no longer be claimed. Keep the subscriber healthy and resolve reported persistence
+errors before accepting further payments.
```

ldk-server-grpc/src/events.rs
```diff
@@ -213,33 +226,41 @@ pub struct PaymentSuccessful {
...
 	/// recover it as LDK Node does not currently persist the failure reason in
 	/// `PaymentDetails`.
-	#[prost(enumeration = "PaymentFailureReason", optional, tag = "2")]
+	#[prost(enumeration = "PaymentFailureReason", optional, tag = "3")]
 	pub reason: ::core::option::Option<i32>,
 }
 /// PaymentClaimable indicates a payment has arrived and is waiting to be manually claimed or failed.
 /// This event is only emitted for payments created via `Bolt11ReceiveForHash`.
+/// Handle every event by its payment ID before `claim_deadline`.
+/// The same invoice can produce more than one event. Fail unexpected duplicate or late payments.
```
tankyleo, line 246: Came up with this here, worth applying to the proto definition too I think:
```diff
 /// This event is only emitted for payments created via `Bolt11ReceiveForHash`.
 /// Handle every event by its payment ID before `claim_deadline`.
 /// The same invoice can produce more than one event. Fail unexpected duplicate or late payments.
+/// Delivery through `SubscribeEvents` is best-effort and is not replayed. If the event is missed and
+/// the payment is not otherwise claimed or failed, LDK Node automatically fails the HTLC backward at
+/// `claim_deadline`.
```

ldk-server-grpc/src/api.rs
tankyleo, line 1: In this file for `pub struct SubscribeEventsRequest {}` this diff below would pair with the documentation update I mentioned above. We'd also apply this to the documentation in the proto defintion at `message SubscribeEventsRequest {}`
```diff
-/// Subscribe to a stream of server events.
+/// Subscribe to a best-effort stream of new server events.
+///
+/// Events are not persisted for subscribers or replayed after reconnecting, and the server does not
+/// wait for client acknowledgement. Slow or disconnected subscribers may miss events. Reconcile
+/// recoverable state with the listing and detail APIs after reconnecting.
+///
+/// If a PaymentClaimable event is missed and the payment is not otherwise claimed or failed, LDK
+/// Node automatically fails the HTLC backward at its claim_deadline.
```

ldk-server-grpc/src/types.rs
tankyleo, line 1: For `struct ForwardedPayment`, seems like `outbound_amount_forwarded_msat` does not need to be optional ?
```diff
-       #[prost(uint64, optional, tag = "4")]
-       pub outbound_amount_forwarded_msat: ::core::option::Option<u64>,
+       #[prost(uint64, tag = "4")]
+       pub outbound_amount_forwarded_msat: u64,
```

ldk-server/src/main.rs
```diff
@@ -610,7 +616,7 @@ fn main() {
 								total_fee_earned_msat,
 								skimmed_fee_msat,
 								claim_from_onchain_tx,
-								outbound_amount_forwarded_msat
+								Some(outbound_amount_forwarded_msat),
```
tankyleo, line 622: When we get rid of the option in `struct ForwardedPayment`, we can also remove the option here

ldk-server/src/util/proto_adapter.rs
```diff
@@ -198,7 +207,7 @@ pub(crate) fn payment_kind_to_proto(
 				status: Some(confirmation_status_to_proto(status)),
 			})),
 		},
-		PaymentKind::Bolt11 { hash, preimage, secret, counterparty_skimmed_fee_msat } => {
+		PaymentKind::Bolt11 { hash, preimage, secret, counterparty_skimmed_fee_msat, .. } => {
```
tankyleo: nit: seems we don't really need these ellipses here and elsewhere in this file ?

ldk-server/src/util/proto_adapter.rs
tankyleo: nit: drop the elipsis here,  replace with `_tx_type` ?

ldk-server/src/util/entropy.rs
```diff
@@ -34,7 +34,7 @@ pub(crate) fn load_or_generate_node_entropy(storage_dir: &Path) -> io::Result<No
 			if let Some(parent) = mnemonic_path.parent() {
 				create_dir_all_private(parent)?;
 			}
-			let mnemonic = generate_entropy_mnemonic(None);
+			let mnemonic = Mnemonic::generate(24).map_err(io::Error::other)?;
```
tankyleo, line 37: Looks like we are back to using `ThreadRng` here instead of `OsRng`, we should ask `rust-bip39` to make a release before we ship this, cc @tnull

ldk-server-grpc/src/types.rs
tankyleo, line 1: Let's expose `amount_msat` in HtlcLocator ?
```diff
 /// Identifies the channel and counterparty that an HTLC was processed with.
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[cfg_attr(feature = "serde", serde(rename_all = "snake_case"))]
 #[cfg_attr(feature = "serde", serde(default))]
 #[allow(clippy::derive_partial_eq_without_eq)]
 #[derive(Clone, PartialEq, ::prost::Message)]
 pub struct HtlcLocator {
        /// The channel that the HTLC was sent or received on.
        #[prost(string, tag = "1")]
        pub channel_id: ::prost::alloc::string::String,
+       /// The amount, in milli-satoshis, of the HTLC that was sent or received.
+       /// This can be unset for events serialized by LDK Node v0.7.0 and prior.
+       #[prost(uint64, optional, tag = "2")]
+       pub amount_msat: ::core::option::Option<u64>,
        /// The `user_channel_id` for the channel.
        /// This can be unset for older serialized events or if the payment was settled on-chain.
-       #[prost(string, optional, tag = "2")]
+       #[prost(string, optional, tag = "3")]
        pub user_channel_id: ::core::option::Option<::prost::alloc::string::String>,
        /// The node id of the counterparty for this HTLC.
        /// This can be unset for older serialized events.
-       #[prost(string, optional, tag = "3")]
+       #[prost(string, optional, tag = "4")]
        pub node_id: ::core::option::Option<::prost::alloc::string::String>,
 }
```

ldk-server/src/main.rs
```diff
@@ -498,77 +502,79 @@ fn main() {
...
 								&event_node,
-								&event_sender,
-								Arc::clone(&paginated_store));
+								&event_sender);
 
-							if let Some(metrics) = &metrics {
-								metrics.update_payments_count(false);
-							}
 						},
 						Event::PaymentClaimable { payment_id, custom_records, claim_deadline, .. } => {
```
tankyleo: Let's forward the `claimable_amount_msat` to our subscribers here ?

ldk-server/src/api/bolt11_claim_for_id.rs
```diff
@@ -29,21 +30,8 @@ pub(crate) async fn handle_bolt11_claim_for_hash_request(
...
-				InvalidRequestError,
-				"Invalid payment_hash, must be a 32-byte hex string.".to_string(),
-			)
-		})?;
-		PaymentHash(hash_bytes)
-	} else {
-		PaymentHash(sha256::Hash::hash(&preimage.0).to_byte_array())
-	};
-
 	let claimable_amount_msat = request.claimable_amount_msat.unwrap_or(u64::MAX);
```
tankyleo, line 33: Found this a little weird, perhaps it's a fix for ldk-node: when the user sets this field, and sets this field to an amount greater than the amount actually claimable, we go ahead and claim the htlc. Shouldn't we scream here instead ? ie you tried to claim x sats, but only y sats were available to claim.

would you consider making this field required instead of optional ? or perhaps we should drop the amount from `claim_for_id` entirely and have the user do the validation of the amount themselves, and decide themselves whether to fail or claim ?

ldk-server/src/main.rs
```diff
@@ -748,22 +754,43 @@ fn main() {
...
+			let payment = payment_to_proto(payment_details);
+
+			let event = payment_to_event(payment);
+			if let Err(e) = event_sender.send(EventEnvelope { event: Some(event) }) {
+				debug!("No event subscribers connected, skipping event: {e}");
+			}
+
+			if let Err(e) = event_node.event_handled() {
+				error!("Failed to mark event as handled: {e}");
+			}
```
tankyleo: nit: can we consolidate all these `if let` statements into a single `if let` after the `match` ?

ldk-server/src/util/metrics.rs
```diff
@@ -91,25 +114,26 @@ impl Metrics {
...
+
+		let public_channels_count =
+			all_channels.iter().filter(|channel_details| channel_details.is_announced).count()
+				as i64;
+		self.total_public_channels_count.store(public_channels_count, Ordering::Relaxed);
+
+		let private_channels_count =
+			all_channels.iter().filter(|channel_details| !channel_details.is_announced).count()
+				as i64;
+		self.total_private_channels_count.store(private_channels_count, Ordering::Relaxed);
```
tankyleo: Let's put these into a helper method, DRY with `update_all_pollable_metrics`. Can also derive private count from channel count - public count so we avoid a traversal.

tankyleo (commented):
One more comment, found with a quick "look for ways to simplify the code" with codex.

ldk-server-grpc/src/types.rs
benthecarman: I just haven't been touching the `ForwardedPayment` struct until https://github.com/lightningdevkit/ldk-node/pull/772 is merged

docs/api-guide.md
benthecarman: Thanks, yeah this is much better

docs/api-guide.md
benthecarman: added

ldk-server-grpc/src/events.rs
benthecarman: added

ldk-server-grpc/src/api.rs
benthecarman: added

ldk-server/src/main.rs
benthecarman: same as above

ldk-server/src/util/proto_adapter.rs
benthecarman: removed

ldk-server/src/util/proto_adapter.rs
benthecarman: #271 was handling this, merged and rebased on top

ldk-server/src/util/entropy.rs
benthecarman: I'll leave as is for now

ldk-server-grpc/src/types.rs
benthecarman: added

ldk-server/src/main.rs
benthecarman: added

ldk-server/src/api/bolt11_claim_for_id.rs
benthecarman: The amount is mostly just a sanity check here. I agree the API is a little weird but it can be a good check to make sure you're backend accounting for whatever you're using hodl invoices for is correct. I had astra improve the documentation here. Lmk what you think

ldk-server/src/main.rs
benthecarman: done

ldk-server/src/util/metrics.rs
benthecarman: done

ldk-server/src/api/bolt11_claim_for_id.rs
tankyleo: Sounds good for now thanks

ldk-server-grpc/tests/htlc_locator_compatibility.rs
```diff
@@ -0,0 +1,41 @@
...
+// This file is licensed under the Apache License, Version 2.0 <LICENSE-APACHE
+// or http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your option.
+// You may not use this file except in accordance with one or both of these
+// licenses.
+
+use ldk_server_grpc::types::HtlcLocator;
+use prost::Message;
+
+// Encoded with the original schema: channel_id = 1, user_channel_id = 2, node_id = 3.
```
tankyleo: Keep telling my AIs that we can aggressively break compat here since we haven't released yet no ?

tankyleo (commented):
Good to squash thanks, also filed this issue in ldk-node to get clarification on this `claimable_amount_msat` parameter

ldk-server-grpc/tests/htlc_locator_compatibility.rs
benthecarman: oh woops, missed that, will remove, maybe should put in the claude.md for now, but we are so close to release so..

benthecarman:
squashed and removed the backwards compat thing

https://github.com/benthecarman/ldk-server/compare/9e77526f9ab3a34d1c134f95f819a40359882612..738caff998a4c5e472760d6aeb8687c839256691

tankyleo (commented):
Worth breaking this PR into a few more commits I think ? No need for any changes to the contents. The comment below is from codex:
```
  > Could you split this PR into more focused commits? 5ceee90 does considerably more than update the ldk-node dependency: it also changes the hold-invoice RPCs from hash to ID, changes payment-event schemas, renames a public
  > payment field, exposes new channel/HTLC fields, and rewrites payment metrics. Likewise, 738caff combines the payment-storage migration with pagination API changes, event-acknowledgement behavior, new PaymentClaimable
  > data, and unrelated CLI/metrics refactors.
  >
  > These are independently reviewable changes_several affecting public API or wire compatibility_and each should have a commit message explaining its intent. At minimum, please separate the dependency/API adaptation, newly
  > exposed fields, payment storage and pagination migration, event-delivery behavior, and cleanup refactors. The protobuf tag compatibility fix should be kept with the isolated payment-event schema commit.
```

benthecarman:
Split up. No changes made https://github.com/benthecarman/ldk-server/compare/738caff998a4c5e472760d6aeb8687c839256691..54a6192ac24f910ba2be8a7d445d620c3eb1a9a0

tankyleo (approved):
thank you!

PR #259, opened 2026-08-19
Expose with-all channel funding variants

tankyleo:
```
    Expose with-all channel funding variants

    Accept 'all' in the CLI and MCP funding amount fields for channel
    opening and splice-in. Represent each choice in gRPC with a typed
    protobuf oneof.

    Dispatch announced, unannounced, and zero-reserve channel opens plus
    splice-in to the corresponding ldk-node with-all methods.

    Co-Authored-By: HAL 9000
```

### DIFF
--- a/ldk-server-cli/src/main.rs
+++ b/ldk-server-cli/src/main.rs
@@ -23,10 +23,10 @@ use ldk_server_client::error::LdkServerErrorCode::{
 	AuthError, InternalError, InternalServerError, InvalidRequestError, LightningError,
 };
 use ldk_server_client::ldk_server_grpc::api::{
-	open_channel_request, splice_in_request, AllFunds, Bolt11ClaimForHashRequest,
-	Bolt11ClaimForHashResponse, Bolt11FailForHashRequest, Bolt11FailForHashResponse,
-	Bolt11ReceiveForHashRequest, Bolt11ReceiveForHashResponse, Bolt11ReceiveRequest,
-	Bolt11ReceiveResponse, Bolt11ReceiveVariableAmountViaJitChannelRequest,
+	onchain_send_request, open_channel_request, splice_in_request, AllFunds,
+	Bolt11ClaimForHashRequest, Bolt11ClaimForHashResponse, Bolt11FailForHashRequest,
+	Bolt11FailForHashResponse, Bolt11ReceiveForHashRequest, Bolt11ReceiveForHashResponse,
+	Bolt11ReceiveRequest, Bolt11ReceiveResponse, Bolt11ReceiveVariableAmountViaJitChannelRequest,
 	Bolt11ReceiveVariableAmountViaJitChannelResponse, Bolt11ReceiveViaJitChannelRequest,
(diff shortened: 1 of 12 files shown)


ldk-server-mcp/src/tools/handlers.rs
```diff
@@ -422,8 +450,46 @@ pub async fn handle_graph_get_node(
...
+			Some(open_channel_request::FundingAmount::AllFunds(_))
+		));
+	}
+
+	#[test]
+	fn parse_request_with_funding_amount_preserves_exact_amount() {
+		let request: SpliceInRequest = parse_request_with_funding_amount(
+			json!({
+				"user_channel_id": "42",
+				"counterparty_node_id": "02example",
```
benthecarman: can we use a real pk here

ldk-server-mcp/src/tools/handlers.rs
```diff
@@ -422,8 +450,46 @@ pub async fn handle_graph_get_node(
...
+				"splice_amount_sats": 50_000
```
benthecarman: can you make this a variable so we can use across both items

ldk-server/src/api/open_channel.rs
```diff
@@ -33,36 +34,66 @@ pub(crate) async fn handle_open_channel(
...
+			(false, false, FundingAmount::ChannelAmountSats(amount_sats)) => {
+				context.node.open_channel(
+					node_id,
+					address,
+					amount_sats,
+					request.push_to_counterparty_msat,
+					channel_config,
+				)?
+			},
+			(true, true, _) => unreachable!("validated above"),
```
benthecarman: why dont we remove the if statement above and just throw the error here now

ldk-server-cli/src/main.rs
benthecarman, line 1: For on-chain we have an amount and a send-all flag. https://github.com/lightningdevkit/ldk-server/blob/main/ldk-server-cli/src/main.rs#L125

Can we switch to do the `all` for the amount instead like you did for the others here

benthecarman (changes requested):

ldk-server-mcp/src/tools/handlers.rs
tankyleo: sorry I missed this fixing this now

tankyleo: done

ldk-server-cli/src/main.rs
tankyleo: done below, current approach breaks our existing users

ldk-server-mcp/src/tools/handlers.rs
tankyleo: done

ldk-server/src/api/open_channel.rs
tankyleo: done

ldk-server-cli/src/main.rs
benthecarman: they are all in this thread 😂

benthecarman (approved):
thanks!

PR #260, opened 2026-08-20
Add Nix deployment support

benthecarman:
ix-bitcoin is archived and no longer receives security fixes. Provide
a maintained upstream path for reproducible builds and declarative
NixOS deployment.

The flake pins Nixpkgs and exports a package and NixOS module. The
package builds the server and CLI, takes its version from Cargo, and
installs shell completions. The module creates a service, user,
configuration, state directory, and optional firewall ports for each
named instance. The documentation covers local builds, deployment,
secrets, and multiple instances.

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -1,0 +1,21 @@
+name: Nix Checks
+
+on: [ push, pull_request ]
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout source code
+        uses: actions/checkout@v6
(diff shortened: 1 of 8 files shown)


tnull (commented):
So far we took the stance that we shouldn't add distro or build system specific code to the repo (mod. docker files, maybe) (see https://github.com/lightningdevkit/ldk-server/pull/68#pullrequestreview-3061288850 for instance). Adding such files will add to the maintenance burden so we should be careful what we commit to supporting.

Do you think it makes sense to officially support this? Or should it just be a separate repo you maintain?

benthecarman:
> Do you think it makes sense to officially support this? Or should it just be a separate repo you maintain?

Yeah with the end of nix-bitcoin I feel it'd be nice if someone does this and me doing separate vs officially supporting feels the same but if is easier I can do it separately

benthecarman:
Thinking about this more I think actually probably better if this just done as a side project.

PR #261, opened 2026-08-25
Add docs to run ldk-server against pruned bitcoind

tankyleo:


docs/pruned-bitcoind.md
```diff
@@ -0,0 +1,124 @@
...
+The persisted-tip retention rule becomes especially important after channels exist. If LDK Server
+cannot obtain every block since its last persisted tip, do not resume routine channel operations
+while it is retrying synchronization. Restore the missing history from an archival backend, allow
+listener synchronization to finish, and only then resume normal operation.
+
+## Wallet Rescans and Recovery
+
+Use `--rescan-from-height` when a fresh wallet must discover transactions created before its first
+startup, for example during mnemonic recovery. Replace the value below with a mainnet height at or
+before the earliest relevant wallet transaction:
```
benthecarman, line 91: According to claude we start scanning at above the given height so need to update. Maybe this is worth fixing in ldk-node though?

https://github.com/lightningdevkit/ldk-node/blob/056447c28221be02c3d39f8c6ae430a67ebbd850/src/builder.rs#L1711-L1721
```suggestion
startup, for example during mnemonic recovery. Replace the value below with a mainnet height
before the earliest relevant wallet transaction:
```

docs/pruned-bitcoind.md
```diff
@@ -0,0 +1,124 @@
...
+If that succeeds, start the initial scan:
+
+```bash
+ldk-server /etc/ldk-server/config.toml --rescan-from-height "$rescan_height"
+```
+
+If the requested range is pruned, LDK Server may expose gRPC while chain synchronization logs
+repeated transient errors. Stop it and recover against an archival backend (or re-download Core's
+chain with pruning disabled). Retry from clean recovery state derived from verified backups; a
+failed first scan may already have persisted its starting checkpoint.
```
benthecarman, line 120: `docs/pruned-bitcoind.md:120` — Recovery advice after a failed first scan is wrong. The BDK wallet is persisted _before_ the rescan checkpoint resolves (`builder.rs:1652-1660` vs tip check at 1670). If the first build fails (height above tip, pruned block, process dies), the next start loads the existing wallet, silently ignores `--rescan-from-height`, and syncs from genesis against the pruned node forever. The doc must say: delete the fresh data directory before retrying.

docs/pruned-bitcoind.md
```diff
@@ -0,0 +1,124 @@
...
+initial chain synchronization finishes.
+
+## Retention and Routine Operation
+
+Keep Bitcoin Core reachable while LDK Server is running. Stop LDK Server gracefully before Core
+maintenance so its latest chain state is persisted. After startup, compare
+`get-node-info.current_best_block.height` with Core's `blocks` value from `getblockchaininfo`.
+
+If LDK Server is offline while Core prunes past its persisted tip, catch-up will fail with
+`Block not available (pruned data)`. A larger prune target reduces this risk but does not express a
```
benthecarman, line 69:
```suggestion
If LDK Server is offline while Core prunes past its persisted tip, catch-up will stall with
`Block not available (pruned data)`. A larger prune target reduces this risk but does not express a
```

this doesn't crash it

docs/getting-started.md
```diff
@@ -58,7 +58,8 @@ rpc_password = "pass"
 ```
 
 Everything else has sensible defaults. See [Configuration](configuration.md) for the full
-reference.
+reference. For a mainnet Bitcoin Core deployment that uses pruning, also review
+[Running with Pruned Bitcoin Core](pruned-bitcoind.md) before choosing its retention target.
```
benthecarman, line 62: > retention target

Can we say prune height or something. I don't think I've ever heard someone call it that

benthecarman (changes requested):

docs/pruned-bitcoind.md
tankyleo: Opened this PR let me know if you think it feels like an improvement, seems so to me

https://github.com/lightningdevkit/ldk-node/pull/1069

PR #262, opened 2026-08-27
Expose BOLT 12 refunds over gRPC

benthecarman:
Clients need both parts of the BOLT 12 refund flow. Add RPCs for creating a refund and requesting its payment, and expose them through the Rust client and CLI.

Add end-to-end coverage for the reverse payment flow and document the new endpoints.


ldk-server/src/api/bolt12_refund.rs
```diff
@@ -0,0 +1,47 @@
...
+
+pub(crate) async fn handle_bolt12_request_refund_request(
+	context: Arc<Context>, request: Bolt12RequestRefundRequest,
+) -> Result<Bolt12RequestRefundResponse, LdkServerError> {
+	let refund =
+		Refund::from_str(&request.refund).map_err(|_| ldk_node::NodeError::InvalidRefund)?;
+	let invoice = context.node.bolt12_payment().request_refund_payment(&refund)?;
+	let payment_id = invoice.payment_hash().to_string();
+
+	Ok(Bolt12RequestRefundResponse { payment_id })
```
jkczyz: Hmmm... we can't do this unfortunately as LDK Node was updated to decouple `PaymentId` from `PaymentHash` in https://github.com/lightningdevkit/ldk-node/pull/948. Even more unfortunately, we rely on LDK for generating the `PaymentId` when the payment is received, so LDK Node will only know it when it processes the `PaymentClaimable` event.

We'll likely want to model `Bolt11ReceiveResponse`, which returns the `PaymentHash`, since they are both inbound payments. So in practice we just need to rename the field in the response and update the docs.

ldk-server/src/api/bolt12_refund.rs
```diff
@@ -0,0 +1,47 @@
...
+use crate::api::error::LdkServerError;
+use crate::service::Context;
+
+pub(crate) async fn handle_bolt12_initiate_refund_request(
+	context: Arc<Context>, request: Bolt12InitiateRefundRequest,
+) -> Result<Bolt12InitiateRefundResponse, LdkServerError> {
+	let route_parameters = build_route_parameters_config_from_proto(request.route_parameters)?;
+	let refund = context.node.bolt12_payment().initiate_refund(
+		request.amount_msat,
+		request.expiry_secs,
```
jkczyz, line 40: Should we reject a 0-expiry? The CLI will give a more reasonable default if left unset, but other clients may forget to set it.

docs/api-guide.md
```diff
@@ -126,12 +126,14 @@ when the invoice is paid.
...
-|-----------------|-------------------------------------------------------------------------|
-| `Bolt12Receive` | Create a BOLT12 offer (fixed or variable amount)                        |
-| `Bolt12Send`    | Pay a BOLT12 offer (with optional quantity, payer note, routing config) |
+### BOLT12 Offers and Refunds
+
+| RPC                    | Description                                                             |
+|------------------------|-------------------------------------------------------------------------|
+| `Bolt12Receive`        | Create a BOLT12 offer (fixed or variable amount)                        |
+| `Bolt12Send`           | Pay a BOLT12 offer (with optional quantity, payer note, routing config) |
+| `Bolt12InitiateRefund` | Create a BOLT12 refund                                                   |
```
jkczyz: Extra space

ldk-server-grpc/src/proto/api.proto
```diff
@@ -379,6 +379,50 @@ message Bolt12SendResponse {
...
+  // Configuration options for payment routing and pathfinding.
+  optional types.RouteParametersConfig route_parameters = 5;
+}
+
+// The response for the `Bolt12InitiateRefund` RPC. On failure, a gRPC error status is returned.
+message Bolt12InitiateRefundResponse {
+
+  // A BOLT12 refund that the recipient can use to request the refund payment.
+  string refund = 1;
+}
```
jkczyz: Since this is equivalent to an outbound payment, we should have a `payment_id` here, but LDK Node doesn't return it. We'll want to do so upstream, IIUC, so we can return it here.

ldk-server-grpc/src/proto/api.proto
```diff
@@ -973,6 +1017,10 @@ service LightningNode {
   rpc Bolt12Receive(Bolt12ReceiveRequest) returns (Bolt12ReceiveResponse);
   // Send a payment for a BOLT12 offer.
   rpc Bolt12Send(Bolt12SendRequest) returns (Bolt12SendResponse);
+  // Return a BOLT12 refund.
+  rpc Bolt12InitiateRefund(Bolt12InitiateRefundRequest) returns (Bolt12InitiateRefundResponse);
+  // Request payment for a BOLT12 refund.
+  rpc Bolt12RequestRefund(Bolt12RequestRefundRequest) returns (Bolt12RequestRefundResponse);
```
jkczyz: We may want to break with the LDK Node naming convention and call these `Bolt12SendRefund` and `Bolt12ReceiveRefund`, though I don't have a strong opinion. The double "request" is just a bit icky.

jkczyz: Also need MCP handlers for these.

benthecarman: yeah this is better, fixed

benthecarman: done

ldk-server/src/api/bolt12_refund.rs
benthecarman: added

ldk-server/src/api/bolt12_refund.rs
benthecarman: Made `Bolt12ReceiveRefundResponse` return payment_hash

docs/api-guide.md
benthecarman: fixed

ldk-server-grpc/src/proto/api.proto
benthecarman: Yeah, seems more systemic than this PR tho, can defer for now?

ldk-server-mcp/src/tools/schema.rs
```diff
@@ -379,6 +379,47 @@ pub fn bolt12_send_schema() -> Value {
...
 }
 
+pub fn bolt12_send_refund_schema() -> Value {
+	json!({
+		"type": "object",
+		"properties": {
+			"amount_msat": {
+				"type": "integer",
+				"minimum": 1,
+				"description": "Amount in millisatoshis to refund"
```
jkczyz, line 388: Should we do this elsewhere, too, for consistency?

ldk-server-mcp/src/tools/handlers.rs
```diff
@@ -247,6 +247,28 @@ pub async fn handle_bolt12_send(client: &LdkServerClient, args: Value) -> Result
...
+pub async fn handle_bolt12_send_refund(
+	client: &LdkServerClient, args: Value,
+) -> Result<Value, McpError> {
+	let mut request: Bolt12SendRefundRequest =
+		parse_request_with_route_parameters(args, |request: &mut Bolt12SendRefundRequest| {
+			&mut request.route_parameters
+		})?;
+	if request.expiry_secs == 0 {
+		request.expiry_secs = DEFAULT_EXPIRY_SECS;
+	}
```
jkczyz, line 259: The schema as `"minimum": 1,`, so this will never be hit? Probably better to remove the minimum from the schema.

ldk-server-grpc/src/proto/api.proto
```diff
@@ -379,6 +379,51 @@ message Bolt12SendResponse {
...
+// payment from this node.
+// See more:
+// - https://docs.rs/ldk-node/latest/ldk_node/payment/struct.Bolt12Payment.html#method.initiate_refund
+message Bolt12SendRefundRequest {
+
+  // The amount in millisatoshis to refund.
+  uint64 amount_msat = 1;
+
+  // Refund expiry time in seconds.
+  uint32 expiry_secs = 2;
```
jkczyz: Should note that zero is rejected.

e2e-tests/tests/e2e.rs
```diff
@@ -987,6 +988,47 @@ async fn test_cli_bolt12_send() {
...
+	);
+	let refund_str = output["refund"].as_str().unwrap();
+	assert!(refund_str.starts_with("lnr"), "Expected lnr prefix, got: {refund_str}");
+	let refund = Refund::from_str(refund_str).unwrap();
+	assert_eq!(refund.amount_msats(), 5_000_000);
+	assert_eq!(refund.quantity(), Some(1));
+	assert_eq!(refund.payer_note().unwrap().to_string(), "test refund");
+
+	let output = run_cli(&server_a, &["bolt12-receive-refund", refund_str]);
+	assert!(!output["payment_hash"].as_str().unwrap().is_empty());
```
jkczyz: Could we check the payment hash now?

ldk-server-mcp/src/tools/schema.rs
benthecarman: removed so we're consistent wit the rest for now

ldk-server-mcp/src/tools/handlers.rs
benthecarman: removed

ldk-server-grpc/src/proto/api.proto
benthecarman: added

e2e-tests/tests/e2e.rs
benthecarman: added, also added mcp tests too

e2e-tests/tests/mcp.rs
```diff
@@ -7,12 +7,39 @@
...
+		panic!("Event stream ended without matching event");
+	})
+	.await
+	.expect("Timed out waiting for event")
+}
+
+fn tool_result_json(response: &Value) -> Value {
+	let text = response["result"]["content"][0]["text"].as_str().unwrap();
+	serde_json::from_str(text).unwrap()
+}
```
jkczyz, line 24: This could be used in a few other pre-existing places. Consider adding another commit to do so.

e2e-tests/tests/mcp.rs
```diff
@@ -7,12 +7,39 @@
...
+		while let Some(Ok(event)) = events.next_message().await {
+			if event.event.as_ref().is_some_and(&pred) {
+				return event;
+			}
+		}
+		panic!("Event stream ended without matching event");
+	})
+	.await
+	.expect("Timed out waiting for event")
+}
```
jkczyz: This is copied from `e2e.rs`. Could you refactor them into `e2e-tests/src/lib.rs`?

ldk-server-mcp/src/tools/schema.rs
```diff
@@ -379,6 +379,45 @@ pub fn bolt12_send_schema() -> Value {
...
+	json!({
+		"type": "object",
+		"properties": {
+			"amount_msat": {
+				"type": "integer",
+				"description": "Amount in millisatoshis to refund"
+			},
+			"expiry_secs": {
+				"type": "integer",
+				"description": "Refund expiry time in seconds"
```
jkczyz: Could you add "(defaults to 86400 if omitted or 0)" like a similar description in `bolt11_receive_schema`?

benthecarman: added

e2e-tests/tests/mcp.rs
benthecarman: done

e2e-tests/tests/mcp.rs
benthecarman: added

jkczyz (approved):

PR #266, opened 2026-09-01
Harden gRPC transport

benthecarman:
Various things found by project loupe. Nothing critical but all worth doing.

This PR was created with assistance from Codex and Claude Code.

### DIFF
--- a/ldk-server/src/main.rs
+++ b/ldk-server/src/main.rs
@@ -665,7 +665,6 @@ fn main() {
 							);
 							let acceptor = tls_acceptor.clone();
 							runtime.spawn(async move {
-								let _handshake_permit = handshake_permit;
 								match tokio::time::timeout(
 									TLS_HANDSHAKE_TIMEOUT,
 									acceptor.accept(stream),
@@ -673,6 +672,10 @@ fn main() {
 								.await
 								{
 									Ok(Ok(tls_stream)) => {
+										// Only the handshake holds a slot. Holding it for the whole
+										// connection would let an unauthenticated peer block new
+										// connections by keeping established ones idle.
+										drop(handshake_permit);
 										let io_stream = TokioIo::new(tls_stream);
 										let mut builder = http2::Builder::new(TokioExecutor::new());
 										builder.max_concurrent_streams(MAX_CONCURRENT_HTTP2_STREAMS);
--- a/ldk-server/src/main.rs
+++ b/ldk-server/src/main.rs
@@ -58,6 +58,7 @@ use crate::util::{systemd, write_new};
 
 const API_KEY_FILE: &str = "api_key";
 const FULL_VERSION: &str = concat!(env!("CARGO_PKG_VERSION"), " (", env!("GIT_HASH"), ")");
+const MAX_CONCURRENT_HTTP2_STREAMS: u32 = 32;
 const MAX_PENDING_TLS_HANDSHAKES: usize = 64;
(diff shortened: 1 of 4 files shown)


ldk-server/src/main.rs
```diff
@@ -655,14 +665,23 @@ fn main() {
 							);
 							let acceptor = tls_acceptor.clone();
 							runtime.spawn(async move {
-								match acceptor.accept(stream).await {
-									Ok(tls_stream) => {
+								let _handshake_permit = handshake_permit;
```
joostjager: [P1] Release this permit once the TLS handshake completes. Because `_handshake_permit` remains in scope across `serve_connection(...).await`, it is held for the entire HTTP/2 connection. An unauthenticated peer can complete 64 TLS handshakes, keep those connections idle, and cause every subsequent connection to be rejected indefinitely.

ldk-server/src/service.rs
```diff
@@ -260,6 +263,15 @@ impl Service<Request<Incoming>> for NodeService {
 		let shutdown_rx = self.shutdown_rx.clone();
 		let (request_parts, request_body) = req.into_parts();
 		let future: Self::Future = Box::pin(async move {
+			let body_permit = match REQUEST_BODY_SEMAPHORE.try_acquire() {
```
joostjager, line 270: [P1] Add a mandatory server-side deadline for request-body collection. This global permit is held while `read_request_body(...).await` waits without a timeout, and only the presence of `x-auth` has been checked. One unauthenticated client can leave eight HTTP/2 bodies unfinished and force every legitimate RPC to fail with `UNAVAILABLE`.

joostjager (commented):
I raised the risk of hand-rolling previously in [PR #220](https://github.com/lightningdevkit/ldk-server/pull/220#issuecomment-4552626715). This PR reinforces that. I think we should reconsider this path and consider migrating to `tonic`.

benthecarman:
> I raised the risk of hand-rolling previously in [PR #220](https://github.com/lightningdevkit/ldk-server/pull/220#issuecomment-4552626715). This PR reinforces that. I think we should reconsider this path and consider migrating to `tonic`.

I asked claude to do an analysis on this while it would have fixed some of these, some of these small issues were on our auth schema which still operates outside of it

ldk-server/src/main.rs
benthecarman: fixed

ldk-server/src/service.rs
benthecarman: fixed

joostjager: The 30-second timeout bounds each individual stalled body, but does not prevent starvation. An unauthenticated client can continuously open replacement streams as the old ones expire, repeatedly occupying all eight global `try_acquire` slots and keeping legitimate RPCs at `UNAVAILABLE`. Could we prevent one unauthenticated peer from monopolizing this pool, and add a regression test for sustained saturation?

joostjager:
> I asked claude to do an analysis on this while it would have fixed some of these, some of these small issues were on our auth schema which still operates outside of it

Agreed, `tonic` would not remove all of our custom authentication logic. My concern is that the recurring issues in these PRs are mostly in transport framing, status and trailer handling, stream lifecycle, and resource limits, which are separate from the auth schema. Even if auth remains custom, there is still an avoidable maintenance burden on the rest.

PR #267, opened 2026-09-01
Reject empty metrics credentials

benthecarman:
Reject configured empty usernames or passwords so Basic Auth cannot be enabled with a predictable empty credential pair.

This commit was created with assistance from Codex.


TheBlueMatt:
Why?

benthecarman:
loupe flagged this and if a user sets this it is likely a misconfiguration. seems better to just throw an error

TheBlueMatt:
No, loupe is dumb. The client shouldn't be enforcing these kinds of things. Even if we think its marginally better to we don't need a ton of code just to enforce random shit on the client-side that the server can deal with.

PR #268, opened 2026-09-01
Harden Config Parsing / Validation

benthecarman:
Various issues by project loupe where we could silently fallback to a default config or file location in an unexpected way. Better to force the user to be explicit upfront to prevent potential issues.

### DIFF
--- a/ldk-server-client/src/config.rs
+++ b/ldk-server-client/src/config.rs
@@ -175,16 +175,12 @@ pub fn resolve_api_key(
 		},
 		None => "bitcoin".to_string(),
 	};
-	if let Some(dir) = storage_dir(config) {
-		let path = api_key_path_for_storage_dir(dir, &network);
-		if let Some(api_key) = read_api_key(&path)? {
-			return Ok(Some(api_key));
-		}
-	}
-
-	match get_default_api_key_path(&network) {
-		Some(path) => read_api_key(&path),
-		None => Ok(None),
+	match storage_dir(config) {
+		Some(dir) => read_api_key(&api_key_path_for_storage_dir(dir, &network)),
+		None => match get_default_api_key_path(&network) {
+			Some(path) => read_api_key(&path),
+			None => Ok(None),
+		},
 	}
 }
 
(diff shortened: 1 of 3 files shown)


ldk-server-client/src/config.rs
joostjager: When `storage_dir(config)` is `Some`, a failed read no longer falls back to the default data directory. Please update the function documentation and the CLI/MCP error messages accordingly. Currently, a missing key in configured storage tells the user to populate the default path, but doing so cannot resolve the error because that path is never checked.

joostjager (dismissed):
Sensible changes

joostjager (commented):
Seems something went wrong in the rebase? https://github.com/lightningdevkit/ldk-server/pull/268/changes#r3912262959 was addressed, but then reverted again in the second push.

Also it would be helpful for review to get a link with the diff that addresses the comments, excluding the rebase. I am not sure what happened.

benthecarman:
sorry fixed, here is link for easy review

https://github.com/benthecarman/ldk-server/compare/72341f04cc84d6752119d5ec10ec6d0d1024a2e8..0cee20cb89c4fc20b37979099fc4cb095ebcfcbf

ldk-server-client/src/config.rs
```diff
@@ -168,19 +169,24 @@ pub fn resolve_api_key(
...
+///
+/// Uses the configured storage directory when one is set. Uses the OS-specific default data
+/// directory only when no storage directory is configured.
+pub fn resolve_api_key_path(config: Option<&Config>) -> Option<PathBuf> {
 	let network = match config {
-		Some(config) => match config.network() {
-			Ok(network) => network,
-			Err(_) => return Ok(None),
-		},
+		Some(config) => config.network().ok()?,
```
joostjager: Non-blocking: `.ok()?` discards the unsupported-network error, so a typo like `regtset` produces a missing-key diagnostic suggesting an API-key override. Could we preserve that error so users are directed to fix the network instead?

joostjager (dismissed):

ldk-server-client/src/config.rs
benthecarman: yeah this seems important, fixed

benthecarman:
Just the fixed invalid network validation

https://github.com/benthecarman/ldk-server/compare/0cee20cb89c4fc20b37979099fc4cb095ebcfcbf..a7f1bb663983420cc2985fbe2058225c22c90930

joostjager (approved):

PR #274, opened 2026-09-08
Configure multiple LSPs as liquidity sources

Camillarhi:
Exposes LDK Node's multi-LSP support so a server can register several LSPs instead of a single one.

Builds on [lightningdevkit/ldk-node#792](https://github.com/lightningdevkit/ldk-node/pull/792), which replaced the per-protocol builder setters with `add_liquidity_source()` and added bLIP-50 / LSPS0 protocol discovery.

LDK Node's `add_liquidity_source()` registers an LSP generically and discovers its supported protocols (LSPS1/LSPS2) via bLIP-50 / LSPS0. Replace the single `[liquidity.lsps2_client]` section with a `[[liquidity.lsps_client]]` array so several LSPs can be registered. Each entry is passed to `add_liquidity_source()` on startup, after which LSPS2 JIT invoices select the cheapest fee offer across every LSPS2-capable LSP.

Surface the per-LSP `trust_peer_0conf` flag, previously hardcoded to `false`. It is required rather than defaulted, since accepting 0-confirmation channels  from an LSP is a trust decision each operator should make explicitly.

### Breaking change

`[liquidity.lsps2_client]` is no longer accepted. Existing configurations must be migrated to `[[liquidity.lsps_client]]` with an explicit `trust_peer_0conf` value:

### DIFF
--- a/contrib/ldk-server-config.toml
+++ b/contrib/ldk-server-config.toml
@@ -66,14 +66,19 @@ server_url = "ssl://electrum.blockstream.info:50002"  # Electrum endpoint
 [esplora]
 server_url = "https://mempool.space/api"      # Esplora endpoint
 
-# LSPS2 Client Support
-[liquidity.lsps2_client]
-# The public key of the LSPS2 LSP we source just-in-time liquidity from.
+# LSPS Client Support
+# Repeat this section to register several LSPs. Supported protocols are discovered per
+# LSP via bLIP-50 / LSPS0 on startup. Only LSPS2 is supported at present.
+[[liquidity.lsps_client]]
+# The public key of the LSP we source inbound liquidity from.
 node_pubkey = "<lsp node pubkey>"
-# Address to connect to the LSPS2 LSP (IPv4:port, IPv6:port, OnionV3:port, or hostname:port).
+# Address to connect to the LSP (IPv4:port, IPv6:port, OnionV3:port, or hostname:port).
 address = "<lsp ip address>:9735"
 # Optional token for authenticating to the LSP.
 # token = ""
+# Accept 0-confirmation channels opened by this LSP. Required for JIT channels to be
+# usable before the funding transaction confirms.
+trust_peer_0conf = true
 
 # Experimental LSPS2 Service Support
 # CAUTION: LSPS2 support is highly experimental and for testing purposes only.
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -147,14 +147,22 @@ You must configure **exactly one** of the following sections:
 > against the blockchain. This means a malicious peer could flood your node with fake channel
 > announcements, consuming memory and disk. If your node is publicly reachable, use bitcoind.
 
-### `[liquidity.lsps2_client]`
+### `[[liquidity.lsps_client]]`
 
-Connects to an [LSPS2](https://github.com/BitcoinAndLightningLayerSpecs/lsp/blob/main/LSPS2/README.md)
-Liquidity Service Provider for just-in-time (JIT) inbound channel opening. When configured,
-the `Bolt11ReceiveViaJitChannel` and `Bolt11ReceiveVariableAmountViaJitChannel` RPCs become
-available, the LSP will open a channel on the fly when the generated invoice is paid.
+Registers a Liquidity Service Provider to source inbound liquidity from. Repeat the section
+to register several LSPs. Each LSP's supported protocols are discovered on startup via
+[bLIP-50 / LSPS0](https://github.com/lightning/blips/blob/master/blip-0050.md). LDK Server
+currently supports LSPS2 only, so an LSP that does not advertise it is registered but unused.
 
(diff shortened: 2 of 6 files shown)


ldk-server/src/util/config.rs
```diff
@@ -1773,6 +1787,91 @@ mod tests {
 		assert!(load_config(&args_config).is_ok());
 	}
 
+	#[test]
+	fn test_multiple_liquidity_sources_from_file() {
```
benthecarman, line 1808: could we also add a test for multiple but one of the lsp configs is invalid, make sure we don't just silently drop the config and properly throw an error

contrib/ldk-server-config.toml
```diff
@@ -66,14 +66,19 @@ server_url = "ssl://electrum.blockstream.info:50002"  # Electrum endpoint
 [esplora]
 server_url = "https://mempool.space/api"      # Esplora endpoint
 
-# LSPS2 Client Support
-[liquidity.lsps2_client]
-# The public key of the LSPS2 LSP we source just-in-time liquidity from.
+# LSPS Client Support
+# Repeat this section to register several LSPs. Supported protocols (LSPS1/LSPS2) are
```
benthecarman: We only support LSPS2 for now

contrib/ldk-server-config.toml
```diff
@@ -66,14 +66,19 @@ server_url = "ssl://electrum.blockstream.info:50002"  # Electrum endpoint
...
+# The public key of the LSP we source inbound liquidity from.
 node_pubkey = "<lsp node pubkey>"
-# Address to connect to the LSPS2 LSP (IPv4:port, IPv6:port, OnionV3:port, or hostname:port).
+# Address to connect to the LSP (IPv4:port, IPv6:port, OnionV3:port, or hostname:port).
 address = "<lsp ip address>:9735"
 # Optional token for authenticating to the LSP.
 # token = ""
+# Accept 0-confirmation channels opened by this LSP. Required for JIT channels to be
+# usable before the funding transaction confirms.
+trust_peer_0conf = false
```
benthecarman: we should set this to `true`, otherwise the UX of JIT channels kinda sucks

docs/configuration.md
```diff
@@ -147,14 +147,21 @@ You must configure **exactly one** of the following sections:
...
+### `[[liquidity.lsps_client]]`
 
-Connects to an [LSPS2](https://github.com/BitcoinAndLightningLayerSpecs/lsp/blob/main/LSPS2/README.md)
-Liquidity Service Provider for just-in-time (JIT) inbound channel opening. When configured,
-the `Bolt11ReceiveViaJitChannel` and `Bolt11ReceiveVariableAmountViaJitChannel` RPCs become
-available, the LSP will open a channel on the fly when the generated invoice is paid.
+Registers a Liquidity Service Provider to source inbound liquidity from. Repeat the section
+to register several LSPs. Each LSP's supported protocols are discovered on startup via
+[bLIP-50 / LSPS0](https://github.com/lightning/blips/blob/master/blip-0050.md), so a single
+entry covers both LSPS1 and LSPS2 depending on what the LSP advertises.
```
benthecarman: LSPS2 only

ldk-server/src/util/config.rs
```diff
@@ -1773,6 +1787,91 @@ mod tests {
...
+			lsps2_service_config_for_feature()
+		);
+
+		fs::write(storage_path.join(config_file_name), toml_config).unwrap();
+		let config = load_config(&args_config).unwrap();
+
+		let lsps_clients = config.lsps_client_config.expect("liquidity sources configured");
+		assert_eq!(lsps_clients.len(), 2);
+
+		assert_eq!(
```
benthecarman, line 1846: instead of individually asserting each field would be cleaner (and more forwards compatible) to create each `LSPSClientConfig` and assert that it at the correct index

ldk-server/src/util/config.rs
```diff
@@ -510,16 +511,18 @@ impl ConfigBuilder {
 		let log_max_files = self.log_max_files.unwrap_or(DEFAULT_LOG_MAX_FILES);
 		let log_to_file = self.log_to_file.unwrap_or(true);
 
-		let lsps2_client_config = self
-			.lsps2
+		let lsps_client_config = self
+			.lsps
```
benthecarman, line 515: For here we would accept duplicate LSPs (same node id), seems we allow that in ldk-node and just silently drop them, should we throw an error here?

benthecarman (changes requested):

ldk-server/src/util/config.rs
Camillarhi: Yes, I have updated it to throw an error here and also included a test for it

ldk-server/src/util/config.rs
Camillarhi: Thanks, this has been updated

benthecarman (approved):

PR #276, opened 2026-09-09
Replace HMAC Auth with scoped macaroons

### LINKED ISSUE #140: Explore more granular auth schemes

benthecarman:
Closes #140 alternative to #273

We initially implemented HMAC auth but it doesn't allow for scoped permissions without creating a new token for each use case. Macaroons support this natively and allows for users to do more complex things without us having to build out tons of support.

This implements macaroons from scratch and replaces our existing HMAC Auth.


docs/api-guide.md
```diff
@@ -15,24 +15,107 @@ underlying LDK Node documentation.
...
+| `node:read` | Node information, balances, and pathfinding scores |
+| `onchain:receive` | Create on-chain receive addresses |
+| `onchain:send` | Send on-chain funds |
+| `invoices:create` | Create BOLT11/BOLT12 invoices and incoming refund requests |
+| `payments:read` | Read payments, forwarded payments, and forwarding statistics |
+| `payments:claim` | Claim or fail held BOLT11 payments |
+| `payments:send` | Send BOLT11, BOLT12, spontaneous, unified, and refund payments |
+| `channels:read` | List channels |
+| `channels:manage` | Open, configure, or cooperatively close channels |
+| `channels:splice` | Splice funds in or out, including to an external address |
```
tnull: Why is `splice` separate from `manage`?

e2e-tests/src/lib.rs
```diff
@@ -139,8 +138,12 @@ pub enum ChainSource {
 		rpc_password: String,
 		rest_address: Option<String>,
 	},
-	Electrum { server_url: String },
-	Esplora { server_url: String },
+	Electrum {
```
tnull, line 149: There are a lot of unrelated formatting included in this jumbocommit. Can we drop them or at least split them out to a separate commit if we want them?

ldk-server-macaroons/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+homepage = "https://lightningdevkit.org/"
+license = "MIT OR Apache-2.0"
+edition = "2021"
+rust-version = "1.85"
+description = "Macaroon credentials and request binding for LDK Server."
+repository = "https://github.com/lightningdevkit/ldk-server/"
+readme = "README.md"
+
+[dependencies]
+ring = { version = "0.17", default-features = false }
```
tnull: Why not `bitcoin-hashes`?

tnull (commented):
Excuse the delay here!

e2e-tests/src/lib.rs
benthecarman: already had https://github.com/lightningdevkit/ldk-server/pull/291 open, removed those from this PR

docs/api-guide.md
benthecarman: Opening and closing channels can't send funds to an attacker (besides fees) whereas splicing out can

ldk-server-macaroons/Cargo.toml
benthecarman: switched in [73cb8d1](https://github.com/lightningdevkit/ldk-server/pull/276/commits/73cb8d1d593e9b1129bb433e6b96526ca54c8c4c)

benthecarman:
split into more logical commits for you

docs/api-guide.md
tnull: Hmm, would it then be more intuitive to have splice-out require `payments:send` permissions and splice in simply `channels:manage`? Or to put differently, I don't see a scenario where  someone want to allow splicing in particular but neither of the other two?

tnull (commented):
LGTM, only open questions would be

a) do we really want to go this way (final decision up to you)
b) do we really need  a separate set of `splice` permissions, given that in my mind they are ~equal to channel management/payment. (see above)

vincenzopalazzo (approved):
LGTM, I think having this allow to share permission with tool like MCP or management tools more easy!

On this line I would be more inclined to keep channels:splice separate. Splice-out can pay an external address. channels manage cannot. Folding splice into manage would let a channel-ops token move funds.

benthecarman:
Split splicing into `splice in` -> `channels:manage` and `splice out` -> `payments:send`. Also fixed a bug where `get-permissions` wouldn't always work

benthecarman:
> do we really want to go this way (final decision up to you)

yeah i think this is better than the original hmac stuff

benthecarman:
added some replay protection after doing a scan. should be good now

tnull (dismissed):
> added some replay protection after doing a scan. should be good now

Hmm, is it worth the added complexity? TLS already gives us replay protection, so the new nonces etc are just added to protect a case where an adversary can somehow still get access to a raw/unencrypted request, no? 

Anyways, non-blocking for me.

benthecarman:
okay sorry for the flip flopping but got convinced on the removing of the nonces/replay

vincenzopalazzo (approved):
LGTM

PR #283, opened 2026-09-14
Use ldk-node for forwarded payment history 

### LINKED ISSUE #162: Add timestamp to `ForwardedPayment`
https://github.com/lightningdevkit/ldk-node/pull/772 will add proper storage for forwarded payments to ldk-node. When we migrate to this, we should add a timestamp to our `ForwardedPayment` return type. It shouldn't be exact so we can't doxx past payments, however the ldk-node PR handles that already.

Found this while working on the thunderhub integration

### LINKED ISSUE #172: Unify database
With ldk-node adding the paginated kv store and allowing us to init the database before the node, we can unify these so any ldk-server specific data can be saved in the same database instead of having 2 db files. Currently we only save paginated versions of the payments and forwarded payments and these will be upstreamed into ldk-node so this doesn't have an immediate usecase but I can imagine us putting auth creds or something in there.

benthecarman:
Closes #162 
Closes #172 

With https://github.com/lightningdevkit/ldk-node/pull/772 merged we can now use this instead of the original ldk-server based version. This also allows us to remove ldk-server's specific sqlite db and implementation.


contrib/ldk-server-config.toml
```diff
@@ -10,6 +10,11 @@ alias = "ldk_server"                          # Lightning node alias
 #async_payments_role = "client"               # Optional async payments role: "client" or "server"
 #enable_zero_fee_commitments = false          # Enable zero-fee commitment channels (default: false)
 
+# Forwarding history: "detailed" retains individual records for the current and previous hour,
+# then aggregates them into hourly channel-pair statistics and removes the individual records.
+# "stats" updates per-channel totals directly without storing individual forwards.
+#forwarded_payment_tracking_mode = "stats"    # Default: "stats"
```
tnull, line 16: Given that this is the default, should we expose `channel_stats`, `list_channel_stats`, and `list_channel_pair_stats` in this PR?

benthecarman: added

tnull (approved):

PR #289, opened 2026-09-19
Postgres follow-ups

tankyleo:
Follow-ups to PR https://github.com/lightningdevkit/ldk-server/pull/243


benthecarman:
Can you add a description

e2e-tests/tests/postgres.rs
```diff
@@ -0,0 +1,598 @@
...
+	wait_for_usable_channel(postgres.client(), &bitcoind, TIMEOUT).await;
+	channels(&sqlite_c, 1).await;
+	let mut before_force_close = Vec::new();
+	for server in [&postgres, &sqlite_c] {
+		let balances = server.client().get_balances(GetBalancesRequest {}).await.unwrap();
+		before_force_close.push((balances, payments(server).await));
+	}
+	// Have SQLite discover the unilateral close onchain, ensuring PostgreSQL's commitment
+	// confirms without a competing commitment broadcast in response to a peer error message.
+	postgres
```
benthecarman, line 530: C can still reconnect to B during the force-close window. The force channel is inbound on C, so ldk-node re-adds B to C's peer store on ChannelPending. B's `disconnect_peer` only clears B's own store. If C's 60s reconnect tick lands before B's commitment confirms, B answers with an error and C force-closes with broadcast, producing a competing commitment. Low probability, but calling `disconnect_peer` from C's side as well removes the risk cheaply.

e2e-tests/tests/postgres.rs
```diff
@@ -0,0 +1,598 @@
...
+	PaymentStatus,
+};
+
+const TIMEOUT: Duration = Duration::from_secs(60);
+
+async fn start_postgres(bitcoind: &TestBitcoind, connection_string: &str) -> LdkServerHandle {
+	let server = LdkServerHandle::start_with_config(bitcoind, |params| {
+		// Each server gets its own table, even when sharing the same test database.
+		let table_name = format!(
+			"node_{}",
```
benthecarman: derives the table name by stripping a leading dot from the tempdir basename. That works only because tempfile uses a `.tmp` prefix and ldk-node splits identifiers on dots. `format!("node_{}", params.grpc_port)` is already unique and obviously valid.

e2e-tests/tests/postgres.rs
```diff
@@ -0,0 +1,598 @@
...
+			server.client().list_channels(ListChannelsRequest {}).await.unwrap().channels;
+		if channels.len() == count && channels.iter().all(|c| c.is_usable) {
+			return channels;
+		}
+		assert!(start.elapsed() < TIMEOUT, "Waiting for {count} usable channels: {channels:?}");
+		tokio::time::sleep(Duration::from_millis(200)).await;
+	}
+}
+
+async fn close_channel(initiator: &LdkServerHandle, peer: &LdkServerHandle, user_channel_id: &str) {
```
benthecarman, line 62: close_channel retries any LightningError for a full 60s in `close_channel`. That masks a real close failure as a generic timeout. Print the first error, or cap the retry to a few seconds.

benthecarman (commented):
Thanks, didn't want to retest anything in ldk-node but these look reasonable.

some claude review

tankyleo:
> didn't want to retest anything in ldk-node but these look reasonable.

thanks yea tried to avoid too much duplication against ldk-node, seems a standalone e2e test in ldk-server could be useful to make sure we don't forget anything.

e2e-tests/tests/postgres.rs
```diff
@@ -0,0 +1,607 @@
...
+			.log(Some("Info"), params.storage_dir.join("ldk-server.log").to_str().unwrap())
+			.build()
+	})
+	.await;
+	assert!(server.storage_dir.join("regtest/ldk_node_postgres.lock").exists());
+	assert!(!server.storage_dir.join("regtest/ldk_node_data.sqlite").exists());
+	server
+}
+
+async fn channels(server: &LdkServerHandle, count: usize) -> Vec<Channel> {
```
benthecarman, line 49: all the helpers in this file might be useful for other tests, can you see if there's any where we can use them and put them in the `lib.rs` instead of just this file

tankyleo: thanks just pushed below to address this

benthecarman:
helpers look great, thanks! two small things claude pointed out about it, then should be good to go

- **Three tests now send or close over gRPC instead of the CLI.** The payment-details, list-payments, and metrics tests use the gRPC send helper, and the channel-lifecycle test uses the gRPC close helper. CLI send and close are still exercised by their own dedicated tests, so nothing is uncovered, but it's a deliberate narrowing worth being aware of.
- **The forwarding-history test got slightly weaker.** It used to assert the detailed record existed immediately after the event, documenting that ldk-node persists before emitting. It now waits up to 30 seconds for it. Fine as a test, but the old comment captured an ordering guarantee that is no longer checked.

for the first might be overkill to do if we already have tests covering but its not too much effort could be worth doing, but just adding back the assert for the second would be nice so we aren't reliant on the helper for doing what we are actually testing

feel free to squash too

tankyleo:
addressed the second point and squashed

.github/workflows/e2e-tests.yml
```diff
@@ -13,6 +13,21 @@ jobs:
   e2e-tests:
     runs-on: ubuntu-latest
 
+    services:
+      postgres:
+        image: postgres:latest
```
benthecarman: i guess github is mandating this now because of all the recent hacks? Can we add the pin or `# zizmor: ignore[unpinned-images]` so we can get a green check

benthecarman (approved):

